PR #1, opened 2021-01-15
Initial setup of repo

rnathuji:
This PR includes changes that layer on top of the state of code from the spike including:

* Switching the project from `yarn` to `npm`
* Restructuring the extension code to live under a `src` directory
* Adding linting
* Adding some helloworld extension tests
* Adding a configuration for CI
* Adding an AGPLv3 LICENSE file

Most of the additions and choices (e.g. of `eslint`, helloworld tests, etc.) are from generated files in a vanilla project created per [these steps](https://code.visualstudio.com/api/get-started/your-first-extension).


.github/workflows/tests.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: tests
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: '0 0 * * 0' # weekly
```
tomjw64, line 9: we are running tests on a schedule and not on PR?

rnathuji: Ah, so this cron line is to run tests weekly on `main` in addition to PRs. For example, you can see runs on this PR [here](https://github.com/openstax/pro-editing-tool/actions), and examples of the scheduling runs [here](https://github.com/openstax/cnx-epub/actions) in `cnx-epub` (which is what I used as a reference when incorporating that).

tomjw64: Oh, i guess the gh actions config file schema is just confusing then...

tomjw64 (approved):

PR #6, opened 2021-01-26
Use webpack for packaging extension code

rnathuji:


tomjw64 (approved):

tomjw64 (commented):
Quick clarification: It seems like we will still dump files compiled with `tsc` into the `out` directory. Is there any reason to keep doing this? (Is it for tests or for extension development?)

rnathuji:
@tomjw64 - The files in `out` are just for the tests. So if you run the `Extension Tests` from the Run menu in vscode, it'll run the `test-compile` script, dump files into `out`, and run tests from there.

PR #20, opened 2021-02-24
Add link validation for open documents

rnathuji:


tomjw64:
Oops. I don't know why I thought this was in the code review column. Sorry!

server/src/utils.ts
```diff
@@ -78,3 +204,93 @@ function generateDiagnostic(severity: DiagnosticSeverity,
...
+      return
+    }
+    const textDocument = request.textDocument
+    let workspaceFolders = await this.connection.workspace.getWorkspaceFolders()
+    if (workspaceFolders == null) {
+      workspaceFolders = []
+    }
+    const diagnostics: Diagnostic[] = []
+    const xmlData = parseXMLString(textDocument)
+    const knownModules = await getCurrentModules(workspaceFolders)
```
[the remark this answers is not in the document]
rnathuji: 🤔  - I think it would require implementing a specific [request](https://microsoft.github.io/language-server-protocol/specification#requestMessage) / response exchange between the two. I haven't really looked at the specifics of using that mechanism, though. But, in general, I do think it's possible based on the protocol spec.

server/src/utils.ts
```diff
@@ -64,6 +65,131 @@ export async function validateImagePaths(textDocument: TextDocument, xmlData: Do
...
+    const [startPosition, endPosition] = calculateElementPositions(linkElement)
+    const preparedDiagnostic: Diagnostic = generateDiagnostic(
+      DiagnosticSeverity.Error,
+      startPosition,
+      endPosition,
+      '',
+      LINK_DIAGNOSTIC_SOURCE
+    )
+
+    if (linkTargetModule == null) {
```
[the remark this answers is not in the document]
rnathuji: Yeah, I thought I had actually added this because of the linter complaining about an explicit `null` check not being there (maybe I had fudged some type thing at the time or something), but removing it now I see it doesn't complain about that anymore. I think it can be a check for empty string instead since the xpath query means the attribute should exist.

server/src/utils.ts
```diff
@@ -64,6 +65,131 @@ export async function validateImagePaths(textDocument: TextDocument, xmlData: Do
...
+      preparedDiagnostic.message = `Target document for link doesn't exist!: ${linkTargetModule}`
+      diagnostics.push(preparedDiagnostic)
+      continue
+    }
+
+    // A matching module was found. Also validate target ID if it's specified in the link
+    const linkTargetId: string = linkElement.getAttribute('target-id')
+    if (linkTargetId !== '') {
+      const targetModuleText = fs.readFileSync(targetModulePath, 'utf-8')
+      const targetModuleXmlData: Document = new DOMParser().parseFromString(targetModuleText)
```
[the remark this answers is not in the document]
rnathuji: So I did see some perf issues, but I noticed it wasn't just due to these other book checks and on some pages even the same page ID validations added up. That's why I added [this change](https://github.com/openstax/poet/pull/20/commits/bee0dc5905fb1e459e1ca4890ded80b01d49e7eb). With that and testing on gitpod, I can type a bunch of text (which would previously have queued a bunch of events and cause a delay for a subsequent introduced error to show up) and still get timely diagnostics on an error. That's not to say memoizing these shouldn't be done, btw, just that the queue approach addresses any end-to-end latency issues.

rnathuji:
@tomjw64 - No worries -- I was going to move it into CR but then wanted to write one more test case thinking it would be quick (but it turns out stubbing some of the things I need is trickier than I initially thought -- or my brain isn't working well at this point in the day 😅 ). Feel free to continue reviewing and providing comments.

server/src/utils.ts
```diff
@@ -64,6 +65,131 @@ export async function validateImagePaths(textDocument: TextDocument, xmlData: Do
...
+export async function validateLinks(xmlData: Document, knownModules: string[]): Promise<Diagnostic[]> {
+  const diagnostics: Diagnostic[] = []
+
+  diagnostics.push(...validateSamePageLinks(xmlData, knownModules))
+  diagnostics.push(...validateOtherPageLinks(xmlData, knownModules))
+
+  return diagnostics
+}
+
+export async function getCurrentModules(workspaceFolders: WorkspaceFolder[]): Promise<string[]> {
```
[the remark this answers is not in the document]
rnathuji: Hmmm...do you mean building the object with title from parsing the CNXML? Perhaps this ties back to the memoization comment as well and might be worth a quick live conversation. I did look at pulling this out of the change event path altogether, but ran into some [issues](https://github.com/openstax/cnx/issues/1356#issuecomment-784461919). Maybe we can talk some tomorrow?

server/src/utils.ts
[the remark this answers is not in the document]
rnathuji: Oh, snap, I think I remember now -- the linter was griping about needing this check when I was experimenting with `xml2js` before switching to `xmldom`. In any case, it's fixed to be a more appropriate check with test coverage now.

server/src/utils.ts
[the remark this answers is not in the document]
tomjw64: Bummer that it'd require a bit of a dance!

server/src/utils.ts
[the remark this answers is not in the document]
rnathuji: I went ahead and [modified](https://github.com/openstax/poet/pull/20/commits/eaa6cc38a4e78eb3d537971ba39a837defd4421e) this function to return a map of ID to `ModuleInformation` object which we can add more stuff to later as / when needed.

server/src/utils.ts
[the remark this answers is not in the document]
rnathuji: In addition to the queue, I pushed changes to incorporate caching in a couple of places (both the [other page XML parsing](https://github.com/openstax/poet/pull/20/commits/1af5c57d6137b370c01708c7b60d82fa246163d7) and [same page valid IDs](https://github.com/openstax/poet/pull/20/commits/a0091152e73a4264d48ddfa58e0b68e056a9dd40)). I also left some general comments with additional context in the [parent issue](https://github.com/openstax/cnx/issues/1356#issuecomment-786142582).

server/src/utils.ts
```diff
@@ -186,7 +187,7 @@ async function validateSamePageLinks(xmlData: Document): Promise<Diagnostic[]> {
       LINK_DIAGNOSTIC_SOURCE
     )
     const linkTargetId: string = linkElement.getAttribute('target-id')
-    if (linkTargetId === '') {
+    if ((linkTargetId === '') || (validTargetIds.has(linkTargetId))) {
```
tomjw64, line 190: What happens if the user deletes a target, but the link is preserved? Will the cache be stale and cause an invalid lack of an error?

server/src/utils.ts
```diff
@@ -135,17 +136,23 @@ async function validateOtherPageLinks(xmlData: Document, knownModules: Map<strin
...
-      const targetModulePath = targetModuleInformation.path
-      const targetModuleText = fs.readFileSync(targetModulePath, 'utf-8')
-      const targetModuleXmlData: Document = new DOMParser().parseFromString(targetModuleText)
-      if (targetModuleXmlData === undefined) {
-        preparedDiagnostic.message = `Could not parse target document!: ${linkTargetModule}`
-        diagnostics.push(preparedDiagnostic)
-        continue
+      // Grab a previously parsed object if we've already checked something on
+      // this target page, otherwise read the file and parse it
+      let maybeModuleXmlData: Document | undefined = parsedOtherPages.get(linkTargetModule)
```
tomjw64, line 141: Will this cache be kept in sync if I am editing the other page?

tomjw64 (commented):
cache correctness questions

server/src/utils.ts
rnathuji: No, because this cache is intra-event and not maintained across events (for some of the reasons mentioned in comments in the parent issue, I think trying to do the latter will lead to various state sync issues). So if you edit / delete a target, errors will show for all links that reference it.

server/src/utils.ts
rnathuji: Yes, so long as the event is fired to validate the first page. For example: If `m0001` has a link to `m00002`, and I modify the target ID / element `m0002` so the reference from `m0001` is bad, then the next time I open / edit `m0001` the error will show. 

Note: The need for the open / edit has nothing to do with the cache (similar to above it is purely intra-event), but with the events we currently hook into for validating a given document. So in terms of cache correctness it's the same as above in that the cache does not attempt to track changes across events.

server/src/utils.ts
tomjw64: Got it. So am I correct that we will only receive a hit in the cache when the same element is targeted by more than one link?

rnathuji: Yes, that is correct.

tomjw64: Neat. Did this help performance with the modules that took a while to process events?

rnathuji: Yeah -- with both changes combined, anecdotally, a page I was using to test (`m58280` from university physics) went from ~1000ms to ~500ms with the combination of both of these (it had 61 links of which 56 were intra-page and 5 other page with IDs).

rnathuji: Or, even better, I can just pull the xpath query [out of the loop altogether](https://github.com/openstax/poet/pull/20/commits/3b011dd2e8bc268bf403a83de128c65e56efefd4) for same page links and build up the valid / duplicate IDs before iterating through the links. Now that same page is down to ~200ms. Should have done that the first time...sigh...

tomjw64 (approved):

PR #22, opened 2021-02-26
Fix bug in element position calculations for diagnostics

rnathuji:
This change addresses the immediate cause of the error which was due to an implicit assumption that elements have siblings (where they can be null). However, the fallback when siblings don't exist is to calculate the end position of the element as just the first few characters of the element using the tag name length. The implication is if a sibling is present the end position captures the entire element, but otherwise it will just capture the tag name. This difference in behavior isn't ideal (see screenshot for an example), but I haven't yet found a simple way to address it. 

Screen Shot 2021-02-26 at 4 20 36 PM

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -259,10 +259,15 @@ export function calculateElementPositions(element: any): Position[] {
   const elementSibling = element.nextSibling
   let endPosition: Position
 
-  // Use the sibling element to get accurate position of the end of the
-  // input element if it's available. Otherwise we'll use the position and
-  // content of the last attribute to establish an end position
-  if (elementSibling === null) {
+  // Establish the end position using as much information as possible
+  // based upon (in order of preference) 1) element sibling 2) final element
+  // attribute 3) the tag
+  if (elementSibling != null) {
+    endPosition = {
+      line: element.nextSibling.lineNumber - 1,
+      character: element.nextSibling.columnNumber - 1
+    }
+  } else if (element.attributes.length > 0) {
     const elementAttributes = element.attributes
     const finalAttribute = elementAttributes[elementAttributes.length - 1]
     const finalAttributeColumn: number = finalAttribute.columnNumber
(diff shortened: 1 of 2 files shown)


rnathuji:
The branch for this fix is based upon [this PR](https://github.com/openstax/poet/pull/20) as the relevant code was modified there. Until it's merged, it may be easier to just look at the [relevant commit](https://github.com/openstax/poet/pull/22/commits/1fa631bb8b98f0abd3cb1414b54064ae79e565c8) directly.

rnathuji:
@Bkblodget - I'd like to get your feedback on the [comment above](https://github.com/openstax/poet/pull/22#issue-581166960), specifically as far as if we're unable to find a way to underline the full element consistently with the red squiggly, would you prefer the heterogeneous behavior demonstrated in the screenshot or a consistent approach that underlines the opening element tag in all cases?

rnathuji:
The question above answered by Bey during our weekly sync: It's preferable to show as much information as possible, so I'll leave this PR as-is for the time being.

rnathuji:
It occurred to me that while @philschatz 's initial suggestion during the team sync of using the attribute lengths to add to the starting position of element start might not work in all cases to check whether attributes themselves are parsed with line / column numbers. It turns out they are, so we can use the position and value of the final attribute to establish an end position as a fallback if no sibling element is present. This makes the difference between the cases pretty minor from a user standpoint:

Screen Shot 2021-03-01 at 1 18 28 PM

I'll push up a commit with this change accordingly.

tomjw64 (approved):

server/src/utils.ts
```diff
@@ -249,16 +249,34 @@ function generateDiagnostic(severity: DiagnosticSeverity,
...
-    character: element.nextSibling.columnNumber - 1
+  const elementSibling = element.nextSibling
+  let endPosition: Position
+
+  // Use the sibling element to get accurate position of the end of the
+  // input element if it's available. Otherwise we'll use the position and
+  // content of the last attribute to establish an end position
+  if (elementSibling === null) {
+    const elementAttributes = element.attributes
+    const finalAttribute = elementAttributes[elementAttributes.length - 1]
```
tomjw64, line 267: Oh wait, what if the element has no next sibling and no attributes?

rnathuji: Ah, good catch. The current callers of this function don't have that happen because of the xpath queries they use, but in general it could. I'll add a test case + fallback to just the tag.

rnathuji: @tomjw64 - [Fixed](https://github.com/openstax/poet/pull/22/commits/d2390a9d195b5e291d515b76de57b74102756743)

tomjw64 (approved):

PR #25, opened 2021-03-08
Add initial CNXML schema validation

rnathuji:
This PR includes the following changes:

* It adds the `redhat.vscode-xml` extension as a dependency in our manifest so that it's automatically installed with ours
* It adds XSD schema files generated using our RelaxNG files (and documentation updates to reflect steps)
* It populates schema files stored in the package to the content repo during activation (this allows us to update schema files with our package while also working with the `xml.catalogs` setting the `vscode-xml` plugin needs to associate schema files with CNXML)

This PR is tied to content repo PRs as well:
* [osbooks-university-physics](https://github.com/openstax/osbooks-university-physics/pull/3)
* [osbooks-college-algebra-bundle](https://github.com/openstax/osbooks-college-algebra-bundle/pull/3)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,3 +110,16 @@ There is a launch configuration to attach to the language server which can be us
 1. Once the extension is launched, execute "Attach to Language Server" from the Run view
 
 You can now set breakpoints, etc. in the server source code.
+
+## Generating XSD schema files
+
+The CNXML schema validation in the extension is performed using XSD files generated using the RelaxNG schema files in the [cnxml repo](https://github.com/openstax/cnxml). The XSD files can be regenerated using [jing-trang](https://github.com/relaxng/jing-trang.git). You can clone that repo and follow the instructions to build `trang.jar` and `jing.jar`. The following steps assume:
+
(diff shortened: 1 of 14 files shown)


tomjw64 (approved):

philschatz:
It seems this PR populates a directory in in the repo, is that correct? And if so, could that directory be gitignored and still work? It would be nice for this plugin to eventually detect a tagging legend and generate XSD files that are specific to the book.

rnathuji:
@philschatz - I'm not sure if this is what you mean by "still work", but yes, the content repos now have a `.gitignore` which includes the populated `.xsd` directory (see [this PR](https://github.com/openstax/osbooks-university-physics/pull/3) and [the corresponding](https://github.com/openstax/osbooks-college-algebra-bundle/pull/3) in `college-algebra-bundle`) and the `vscode-xml` extension still picks up the schema files + works as expected. I believe that addresses your use case for tagging legend too, but LMK if I'm misunderstanding your question / use case.

rnathuji:
Oh neat, I didn't know about that feature. I can spin one up real quick. Any preference / suggestion for a name for that repo?

PR #32, opened 2021-03-19
Simplify npm commands

philschatz:
:art: Webpack ran more often than necessary and it was difficult to jut run the server/client/cypress tests and any prerequisite steps. 

This also provides a `npm run clean` command instead of implicitly cleaning in the pretest shell script.


package.json
```diff
@@ -12,22 +12,23 @@
...
     "cypress:open": "cypress open",
-    "vscode:prepublish": "npm run _webpack_ -- --mode production",
-    "_webpack_": "webpack --stats=errors-warnings --config ./client/webpack.config.js --config ./server/webpack.config.js",
-    "webpack": "npm run _webpack_ -- --mode development",
-    "webpack:watch": "npm run _webpack_ -- --mode development --watch",
     "package": "vsce package",
-    "build": "tsc --build",
-    "watch": "npm run build -- -watch",
+    "build:ts": "tsc --build",
+    "build:webpack": "webpack --stats=errors-warnings --mode development --config ./client/webpack.config.js --config ./server/webpack.config.js",
```
tomjw64, line 19: I know it's maybe a bit out of scope, but could we go with `--stats=minimal`? I've found it super helpful to know when the errors that show are from the latest watched build or not

tomjw64 (commented):
No more production builds?

philschatz:
Production builds are back! `npm run build:production`

tomjw64 (approved):

PR #36, opened 2021-03-30
More XSD schema updates for false positive errors

rnathuji:
This PR contains generated XSD file updates based upon changes to [POET specific RelaxNG schema files](https://github.com/openstax/cnxml/compare/master...poet-schema). See [parent issue](https://github.com/openstax/cnx/issues/1481) for context / details.


philschatz (approved):
:tada: Great! It looks like the only custom namespaces in our books were katalyst and xsl-fo. The xsl-fo attributes can be removed from the content but that's for another story

PR #37, opened 2021-03-30
add TypeScript definitions

therealmarv:
add typescript openstax/cnx#1483


therealmarv:
@philschatz it's ready for review.

About the lines with `// eslint-disable-line @typescript-eslint/...` I could not make eslint and webpack make both happy without a major rewrite although I think the code is already good enough on those lines.

How the npm lint would look like without the two exceptions.
```
  21:14  error  Unexpected object value in conditional. The condition is always true                 @typescript-eslint/strict-boolean-expressions
  59:9   error  Unexpected nullable object value in conditional. An explicit null check is required  @typescript-eslint/strict-boolean-expressions
```

My thoughts:
- Line 21: is clearly not every time true in the loop. This is too restrictive linting in my opinion.
- Line 59: I do a null check already. But it seems it's not explicit enough for Typescript, argh.

But comments are welcome. Normally I don't disagree with a linter.

client/src/webview-js/image-upload/image-upload.ts
```diff
@@ -1,22 +1,30 @@
...
+  const uploadButton = document.getElementById('trigger-upload') as HTMLElement
+  const cancelButton = document.getElementById('cancel-upload') as HTMLElement
 
-  const clearPending = () => {
+  const clearPending = (): void => {
     pending = []
-    while (preview.firstChild) {
-      preview.firstChild.remove()
+    if (preview.firstChild !== null) {
+      while (preview.firstChild) { // eslint-disable-line @typescript-eslint/strict-boolean-expressions
```
tomjw64, line 21: I think the linter probably wants something like:
```suggestion
      while (preview.firstChild != null) {
```
which would also allow you to drop the `if`

client/src/webview-js/image-upload/image-upload.ts
```diff
@@ -45,21 +53,23 @@ window.addEventListener('load', () => {
     }
   }
 
-  const handleDrop = (e) => {
-    const files = e.dataTransfer.files;
+  const handleDrop = (e: DragEvent): void => {
+    const files = e.dataTransfer?.files
 
-    ([...files]).forEach(previewFile)
+    if (files) { // eslint-disable-line @typescript-eslint/strict-boolean-expressions
```
tomjw64, line 59: I think the linter probably wants
```suggestion
    if (files != null) {
```

client/src/webview-js/image-upload/image-upload.ts
```diff
@@ -1,22 +1,30 @@
-let vscode
-let pending = []
+declare let acquireVsCodeApi: any
+let vscode: any
+
+interface PendingEntry {
+  mediaName: string
+  data: string | ArrayBuffer | null
```
tomjw64, line 6: We should only expect data to be a `string`
```suggestion
  data: string
```
because we only use `readDataAsUrl()` on the `FileReader` below, it should be fine for us to cast `reader.result` to `string` when we push onto `pending`. e.g.
```js
pending.push({
  mediaName: file.name,
  data: reader.result as string
})
```

client/src/webview-js/image-upload/image-upload.ts
```diff
@@ -1,22 +1,30 @@
...
   vscode = acquireVsCodeApi() // eslint-disable-line no-undef
 
-  const dropArea = document.getElementById('drop-area')
-  const preview = document.getElementById('preview')
-  const uploadButton = document.getElementById('trigger-upload')
-  const cancelButton = document.getElementById('cancel-upload')
+  const dropArea = document.getElementById('drop-area') as HTMLElement
+  const preview = document.getElementById('preview') as HTMLElement
+  const uploadButton = document.getElementById('trigger-upload') as HTMLElement
+  const cancelButton = document.getElementById('cancel-upload') as HTMLElement
```
tomjw64, line 16: Might be a good use of the `expect` function that's been put into our `utils` files for the client and the server (may need to use the server version, since the client version contains a vscode api call). Here's the function:
```
function expect<T>(value: T | null | undefined, message: string): T {
  if (value == null) {
    throw new Error(message)
  }
  return value
}
```
The gain out of using this is that we'd end up actually checking to see if we can treat these as `HTMLElement` before we start using them and throw an error in the correct place otherwise.
```suggestion
  const dropArea = expect(document.getElementById('drop-area'), 'html file must contain #drop-area')
  const preview = expect(document.getElementById('preview'), 'html file must contain #preview')
  const uploadButton = expect(document.getElementById('trigger-upload'), 'html file must contain #trigger-upload')
  const cancelButton = expect(document.getElementById('cancel-upload'), 'html file must contain #cancel-upload')
```

therealmarv:
super helpful @tomjw64 will rework the parts

tomjw64:
Rebasing/merge with latest on main may solve the coverage issues (it is 100% on main)

tomjw64 (approved):

philschatz (approved):

therealmarv:
On #1425 I will try (again) to move the helper function into an external library. I will leave it as it is for now because I had again problems importing the external function into a test function.

PR #40, opened 2021-04-02
Change refresh panel semantics per suggestion

tomjw64:


philschatz (approved):
Thanks for clarifying the methods

PR #41, opened 2021-04-05
Add validation check when publishing

rnathuji:
In addition to implementing the validation check logic when publishing, this PR updates the `m00001` test content in our repo to pass schema validation so we have one CNXML that is "valid" in the repo.


client/src/push-content.ts
```diff
@@ -1,7 +1,33 @@
...
+  cnxmlErrorMsg: 'There are outstanding validation errors that must be resolved before pushing is allowed.',
+  xmlErrorMsg: 'There are outstanding schema errors. Are you sure you want to push these changes?',
+  xmlErrorIgnoreItem: 'Yes, I know more than you'
+}
+
+export const canPush = async (errorsBySource: Map<string, Array<[vscode.Uri, vscode.Diagnostic]>>): Promise<boolean> => {
+  if (errorsBySource.has(DiagnosticSource.cnxml)) {
+    void vscode.window.showErrorMessage(PushValidationModal.cnxmlErrorMsg, { modal: true })
+    return false
+  } else if (errorsBySource.has(DiagnosticSource.xml)) {
```
tomjw64, line 20: Subjective: it looks like the parent `if` is being used as a guard, so this else if could also be a guard ( via dropping the `else`)

client/src/push-content.ts
```diff
@@ -1,7 +1,33 @@
...
+    const selectedItem = await vscode.window.showErrorMessage(PushValidationModal.xmlErrorMsg, { modal: true }, PushValidationModal.xmlErrorIgnoreItem)
+    if (selectedItem === PushValidationModal.xmlErrorIgnoreItem) {
+      return true
+    } else {
+      return false
+    }
```
tomjw64, line 26:
```suggestion
    return selectedItem === PushValidationModal.xmlErrorIgnoreItem
```

client/src/utils.ts
```diff
@@ -166,3 +166,27 @@ export function launchLanguageServer(context: vscode.ExtensionContext): Language
 
   return client
 }
+
+export function getErrorDiagnosticsBySource(): Map<string, Array<[vscode.Uri, vscode.Diagnostic]>> {
+  const errorsBySource = new Map<string, Array<[vscode.Uri, vscode.Diagnostic]>>()
+  const diagnostics = vscode.languages.getDiagnostics()
+  let uri, fileDiagnostics
+
+  for ([uri, fileDiagnostics] of diagnostics) {
```
tomjw64, line 175: Subjective: Is this syntax possible here, so we don't need to re-bind the same variables every loop?
```suggestion
  for (const [uri, fileDiagnostics] of diagnostics) {
```

tomjw64 (approved):
All my suggestions are pretty much up to subjective interpretation. Looks good as is.

rnathuji:
@tomjw64 , @ckline-tryptic - So I think the CI is griping because [these two lines](https://github.com/openstax/poet/blob/63f89b2ee8ace7a285a49065a4010f64a128592d/client/src/push-content.ts#L36-L37) are currently missed by tests. I've been unable to find a way to mock intra-module dependencies, but I think (currently in the process of trying / confirming) if I move `getRepo()` to `utils.ts` I could then stub it along with everything else to invoke `pushContent()` and get coverage on what's missing. For example:

```javascript
import * as utils from './../../utils' // Used for dependency mocking in tests

test('my new test case', async () => {
  const showErrorMsgStub = sinon.stub(vscode.window, 'showErrorMessage')
  const showInfoMsgStub = sinon.stub(vscode.window, 'showInformationMessage')
  sinon.stub(utils, 'getErrorDiagnosticsBySource').returns(...)
  sinon.stub(utils, 'getRepo').returns(...)
})
```

Do y'all have thoughts on: 1) Moving `getRepo()` to `utils.ts`; 2) Another, simpler option  for how to test this without the reorganization?

rnathuji:
Actually, you can ignore my previous comment. For some reason I thought when I tried stubbing a non-imported function within a module in the past with `sinon` it didn't work, but the following seems to be behaving as expected:

```javascript
test('my new test case', async () => {
  const showErrorMsgStub = sinon.stub(vscode.window, 'showErrorMessage')
  const showInfoMsgStub = sinon.stub(vscode.window, 'showInformationMessage')
  sinon.stub(pushContent, 'canPush').resolves(false)
  sinon.stub(pushContent, 'getRepo').returns(...)
})
```

So I was needlessly overcomplicating things...sigh...

PR #42, opened 2021-04-05
Implement hard refocus to simulate cnxml preview resource rebind

tomjw64:


rnathuji (approved):
This seems to fix the scenario as described in the issue. I did find something kind of odd when playing around where it seems that clicking the preview button versus the keyboard shortcut shift+command+v seem to open separate preview panels which independently update (e.g. if you click preview across files a single tab gets updated, but if you intermingle preview button with the keyboard shortcut it seems to target a separate tab). 

I believe this may be because the command [here](https://github.com/openstax/poet/blob/361eb64ba411c10c02b41849b11edfcff12ba21b/package.json#L129) should be `openstax.showPreviewToSide` 🤷‍♂️ . In any case, I'm mentioning it here for discussion / future reference, but it's not directly related to this PR.

PR #46, opened 2021-04-12
Add ToC trees sidebar view

rnathuji:


client/src/extension.ts
```diff
@@ -67,13 +75,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<(typeo
   }
 
   const lazilyFocusOrOpenPanelOfType = createLazyPanelOpener(activationByType)
+  tocTreesProvider = new TocTreesProvider(client)
 
   vscode.workspace.onDidChangeWorkspaceFolders(ensureCatch(forwardOnDidChangeWorkspaceFolders(client)))
-  client.onRequest('onDidChangeWatchedFiles', ensureCatch(refreshTocPanel(client)))
+  client.onRequest('onDidChangeWatchedFiles', ensureCatch(invokeRefreshers([refreshTocPanel(client), async () => tocTreesProvider.refresh()])))
```
tomjw64, line 81: Since tocTreesProvider is never disposed, would we be able to include this `onDidChangeWatchedFiles` listener in the class itself? I've recently made a [change](https://github.com/openstax/poet/pull/47/files#diff-7b7aa9c8b1944a19471cbe4172196463223080d88b4433c79933f3a47f8ca0f3L72) that removes this listener from the base extension file in favor of each component listening to the events it needs (the listener for the toc Panel was only here before the change because panels can be disposed at any time and we didn't have a mechanism for cleaning up the listener with the panel yet)

client/src/toc-trees.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  async getChildren(element?: TocTreeItem): Promise<TocTreeItem[]> {
+    if (element !== undefined) {
+      return element.children
+    }
+
+    const uri = expect(getRootPathUri(), 'No workspace root for ToC trees')
+    const bundleTrees: BundleTreesResponse = await this.client.sendRequest(
+      ExtensionServerRequest.BundleTrees,
+      { workspaceUri: uri.toString() }
```
tomjw64, line 30: Typescript doesn't end up type checking this argument (since the signature of `sendRequest` allows `any`). One option would be to call a more type-safe wrapper function like [this one](https://github.com/openstax/poet/blob/73200cb45af3f687fbf59e0739d0e0aef1195a62/client/src/panel-toc-editor.ts#L59) (which could be either copied or moved to a shared location and exported)

tomjw64 (approved):

client/src/extension.ts
rnathuji: Yep, can definitely do this. I am hitting something unexpected, though, when adding it:

```javascript
  constructor(private readonly client: LanguageClient) {
    const fileWatchers: vscode.FileSystemWatcher[] = [
      vscode.workspace.createFileSystemWatcher('**/modules/**'),
      vscode.workspace.createFileSystemWatcher('**/collections/**')
    ]
    fileWatchers.forEach(watcher => {
      watcher.onDidChange((e) => this.refresh())
      watcher.onDidCreate((e) => this.refresh())
      watcher.onDidDelete((e) => this.refresh())
    })
  }
```

I get the events as expected in the class, but there seems to be a race condition now where the data returned by the language server isn't always the latest. For example, if I change the module title, the event / request / response are triggered as expected, but the data is sometimes still old. If I manually refresh again the expected data shows up. I believe before the event path flowed through the server [here](https://github.com/openstax/poet/blob/73200cb45af3f687fbf59e0739d0e0aef1195a62/server/src/server.ts#L132) so that may be why this is occurring with the separate watcher since then the file events to the components can be in reverse order. In that case (I think) the server may not return the latest data from the file system if the `BundleTrees` request arrives before it has serviced its own `onDidChangeWatchedFiles` handler. However, if I'm not mistaken and this is the case, it may affect the ToC editor panel the same way so figured I'd ask if you've seen anything like this as well as part of your changes?

tomjw64: I'm not sure I understand. Why make the new file watchers? Wouldn't doing that invite the race condition? Does this not work?
```js
constructor(private readonly client: LanguageClient) {
  client.onRequest('onDidChangeWatchedFiles', this.refresh.bind(this))
}
```

client/src/toc-trees.ts
rnathuji: [Done](https://github.com/openstax/poet/pull/46/commits/f9c31dc4dd67b1821963e76991bb38cbaf5da931)!

client/src/extension.ts
rnathuji: Ah, my mistake, I misinterpreted your initial comment + didn't look at the corresponding changes closely enough. However, does `client.onRequest` would overwrite previous watchers? For example, if I use the code you suggest above, I find that reordering the following lines in `extension.ts`, impacts whether things will work or not (the following works in the TocTrees but seemingly breaks the ToC editor in my branch, reversing them does the opposite):

```javascript
client.onRequest('onDidChangeWatchedFiles', ensureCatch(refreshTocPanel(client)))
tocTreesProvider = new TocTreesProvider(client)
```

I realize that the specific `client.onRequest()` above would be going away in `extension.ts`, but I think we're still sharing a client object in the context of your changes, so can both components invoke `client.onRequest()` w/o affecting each other? Or am I still fudging something here?

tomjw64: Hmm... that is definitely not what I would expect given how the rest of the event driven stuff in the vscode api works. I can take a closer look this afternoon if you don't beat me to it

rnathuji: Just as a brief follow up, I think [this](https://github.com/microsoft/vscode-languageserver-node/blob/d58c00bbf8837b9fd0144924db5e7b1c543d839e/jsonrpc/src/common/connection.ts#L1291) may be the underlying reason for the behavior I mentioned above.

tomjw64: Yikes. Maybe to prevent from using one big undisposable listener or from having one component from owning this request, we could demultiplex via `EventEmitter`. Something similar to the pattern you used for `onDidChangeTreeData`? What do you think?
```ts
// extension.ts
const onDidChangeWatchedFilesEmitter: vscode.EventEmitter<undefined> = new vscode.EventEmitter()
const onDidChangeWatchedFiles = onDidChangeWatchedFilesEmitter.event
client.onRequest('onDidChangeWatchedFiles', () => { onDidChangeWatchedFilesEmitter.fire(undefined) })
// ... pass onDidChangeWatchedFiles to components as part of the extension's context
```

tomjw64: Perhaps this is for me to change as I rebase my scroll syncing branch, though. Seems potentially out of scope for this PR, since it's a bit dependent on work that hasn't been merged yet.

tomjw64 (approved):

client/src/extension.ts
rnathuji: Sounds like a plan on both counts. I'll go ahead and merge this PR, and we can iron this out in the scroll syncing branch.

PR #47, opened 2021-04-12
Add scroll sync & change panel implementation

tomjw64:
Tests and code coverage still need a lot of work, but the features for scroll syncing are there and can be reviewed at a high level


rnathuji:
@tomjw64 - FWIW, I went through this PR at a high-level and didn't have any major questions (TBH I was mostly just trying to get a sense of how the changes worked 😅 ). I know it's still in draft, and I think we still want to massage in the file change event stuff we discussed in [PR 46](https://github.com/openstax/poet/pull/46), but otherwise it seems to work / looks good to me in general 🤷‍♂️ .

A couple of minor things I noticed (both seem to be related to line number rounding / estimation):

* When scrolling in larger files on the editor side (specifically `m49444` in college algebra bundle which is pretty math heavy in case that's useful for reproducing locally), with the preview open there's some glitchiness where it "kicks back" (sorry, there's probably a better way to describe this). I think this may have to do with feedback happening when the preview is scrolled.
* If you scroll down to the end of a CNXML file, the preview will eventually go back to the top. This seems to be because `nextIndex` will get to `-1` [here](https://github.com/openstax/poet/blob/be25564ecb86230cd20fee83dc976a64888dcf03/client/src/webview-js/cnxml-preview/cnxml-preview.js#L126).

tomjw64:
So tests are consistently failing right now due to some issue that seems to only occur on CI, where the cnxml preview panel is not able to bind to any resource at all (even if I try to do explicitly re-bind it in the test to the document in the active editor). I will have to spend some more time investigating this, since there is no obvious answer I can think as to why this would only occur in CI.

tomjw64:
I'm going to enlist Phil's help on these issues tomorrow. For now, I am just documenting the changes made in the cnxml preview webview, since they may be a bit confusing for a reader.

client/src/panel-cnxml-preview.ts
```diff
@@ -1,105 +1,230 @@
...
-
-  async function updatePreview(): Promise<void> {
-    clearTimeout(throttleTimer)
-    if (disposed) {
+  panel.webview.html = rawTextHtml('Loading...')
+  return panel
+}
+
+export const isCnxmlFile = (document: vscode.TextDocument): boolean => {
+  return document.languageId === 'cnxml'
```
rnathuji, line 71: Does this ever return `true`? I believe due to the following in our `settings.json`, for CNXML `languageId` will be `xml`:

```json
"files.associations": {
    "*.cnxml": "xml"
}
```

This wouldn't impact correctness where `isCnxmlFile` is used today since `isXmlFile` will return `true` but  could possibly cause confusion in the future 🤷‍♂️ . 

Just as a note: The association of `*.cnxml` with `xml` in `settings.json` goes back to [this](https://github.com/openstax/cnx/issues/1354#issuecomment-772862496).

rnathuji (approved):

client/src/panel-toc-editor.ts
```diff
@@ -176,8 +156,8 @@ export const refreshPanel = async (panel: vscode.WebviewPanel, client: LanguageC
   const trees = await requestBundleTrees(client, { workspaceUri: uri.toString() })
   const allModules = await requestBundleModules(client, { workspaceUri: uri.toString() })
   const orphanModules = await requestBundleOrphanedModules(client, { workspaceUri: uri.toString() })
+  /* istanbul ignore if */
```
philschatz, line 159: Could this just be `expect(trees, 'Server cannot properly find workspace')`?

client/src/panel.ts
```diff
@@ -0,0 +1,240 @@
...
+ * panel to the host. `OutMessage` is the type of messages sent from the host
+ * to the internal webview panel.
+ */
+export abstract class Panel<InMessage, OutMessage> implements DisposableSupplemental, Messageable<InMessage, OutMessage> {
+  protected readonly panel: vscode.WebviewPanel
+  private readonly disposer: DisposableSupplemental
+
+  constructor(initPanel: () => vscode.WebviewPanel) {
+    this.disposer = new Disposer()
+    this.panel = initPanel()
```
philschatz, line 74: I like having a common base for all panels. If we're immediately calling initPanel in the constructor why not make the constructor take a Panel instead of a function that returns a Panel?

:thinking: Also, (for another PR) since this is a base for all panels, would it make sense to perform all the CSP incantations here in a common place?

client/src/webview-js/cnxml-preview/cnxml-preview.js
```diff
@@ -177,35 +359,23 @@ function __vdom__applyProps($el, newProps, oldProps = {}) {
...
-        __vdom__removeProp($el, name)
+        $el.removeAttribute(name)
       } else if (newValue !== oldValue) {
-        __vdom__setProp($el, name, newValue)
+        $el.setAttribute(name, newValue)
       }
     }
   })
 }
-function __vdom__setProp($el, name, value) {
```
philschatz, line 191: Yeah, our code doesn't add/remove classes and if it did we could just write `'class': ...`

philschatz (approved):
Awesome! Just one question on panel-instantiation

client/src/panel.ts
tomjw64: Huh... I'm not sure why I made it a function that returns a Panel... Must've over-thought it. I'll change that.

As for CSP stuff, I'm a little hesitant to do that, since the Panel class is abstract and really doesn't care at all what inheritors do when it comes to content. CSP isn't always listed out right now, e.g. there's a loading state for cnxml preview, where I didn't bother using csp stuff, and the panel class doesn't stop me. Were you thinking of this being a helper of some sort, or a mandate that the panel class enforces (that all displayed content must specify a CSP)?

client/src/panel-toc-editor.ts
tomjw64: Sure. Less coverage ignoring!

client/src/panel.ts
tomjw64: Let's continue this discussion, but I'm gonna go ahead and merge this PR, since it's got the two approvals, and I've addressed most of the things. I suspect I may have to come back to anti-jitter behavior as well sometime soon as well.

PR #49, opened 2021-04-15
Tagging for release

ckline-tryptic:


client/src/push-content.ts
```diff
@@ -47,14 +75,46 @@ export const getMessage = async (): Promise<string | undefined> => {
...
 
+  const tagging = await _taggingDialog()
+  let tag: string | undefined
+  if (tagging !== 'No Tag') {
+    /* instanbul ignore next */
+    const tagMode = tagging === 'Release'
+    tag = _getTag(repo, tagMode)
+  }
+
+  if (tag !== undefined) {
```
rnathuji, line 101: Since `_getTag` can return `undefined`, is it possible that the user requests a tag but the commit gets pushed without one since we don't check explicitly after L98 and error if `undefined` is returned there like in L107 below?

client/src/push-content.ts
```diff
@@ -47,14 +75,46 @@ export const getMessage = async (): Promise<string | undefined> => {
...
+    /* instanbul ignore next */
+    try {
+      await (repo as any)._repository.tag(tag) // when VSCode API is updated -> await repo.tag(tag)
+    } catch (e) {
+      const message: string = e.gitErrorCode === undefined ? e.message : e.gitErrorCode
+      void errorReporter(`Tagging failed: ${message} ${String(e.stderr)}`)
+    }
+    return
```
rnathuji, line 109: Did you want to return after posting the error message in the line above?

client/src/push-content.ts
```diff
@@ -47,14 +75,46 @@ export const getMessage = async (): Promise<string | undefined> => {
...
+  }
+
+  const diff = await repo.diffWithHEAD()
```
rnathuji, line 112: 🤔 - Is there a case where we'd want to apply a tag but there's no diff? I'm wondering if this check should actually go into `canPush()` (or at least earlier in this flow) if we always want to do nothing when there's no local changes?

rnathuji (commented):
In addition to some of the inline comments, it looks like the `push to new branch` test may be failing in CI.

client/src/push-content.ts
```diff
@@ -32,7 +33,34 @@ export const getRepo = (): Repository => {
...
+  /* instanbul ignore next */
+  return await vscode.window.showInformationMessage(
+    'Are you tagging for release?',
+    // { modal: true },
+    ...['No Tag', 'Release Candidate', 'Release']
+  )
+}
+
+export const getNewTag = (repo: Repository, release: boolean): string | undefined => {
+  /* instanbul ignore next */
```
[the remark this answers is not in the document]
ckline-tryptic: The ignores are here because I separated testing into another card.

client/src/push-content.ts
```diff
@@ -1,6 +1,7 @@
 import vscode from 'vscode'
 import { expect, getErrorDiagnosticsBySource } from './utils'
-import { GitExtension, GitErrorCodes, CommitOptions, Repository } from './git-api/git'
+import { GitExtension, GitErrorCodes, CommitOptions, Repository, RefType } from './git-api/git'
+import { integer } from 'vscode-languageserver-types'
```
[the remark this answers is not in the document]
ckline-tryptic: I guess I figured It made it more clear what was going on, but we can do that.

client/src/panel-toc-editor.ts
```diff
@@ -42,15 +42,15 @@ export interface ModuleRenameSignal {
   moduleid: string
   newName: string
 }
-export type PanelIncomingMessage =
-  (DebugSignal
+export type PanelIncomingMessage = (
```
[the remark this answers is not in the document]
ckline-tryptic: I was going over some stuff with Brenda, and I think something happened when I merged main.

client/src/push-content.ts
```diff
@@ -32,7 +33,34 @@ export const getRepo = (): Repository => {
   return result
 }
 
+export const taggingDialog = async (): Promise<string | undefined> => {
+  /* instanbul ignore next */
+  return await vscode.window.showInformationMessage(
+    'Are you tagging for release?',
+    // { modal: true },
```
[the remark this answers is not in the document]
ckline-tryptic: yep

client/src/push-content.ts
ckline-tryptic: It looks like tagging and committing will stay separate, since a user may commit and then later decide to tag.

client/src/push-content.ts
ckline-tryptic: Tagging and committing will remain separate.

client/src/push-content.ts
```diff
@@ -32,7 +33,34 @@ export const getRepo = (): Repository => {
...
+  // could probaly use a filter-reduce here
+  repo.state.refs.forEach(ref => {
+    if (ref.type !== RefType.Tag) { return }
+    if (release === ref.name?.includes('rc')) { return }
+
+    const versionNumberString = expect(ref.name, '').replace('rc', '')
+    tags.push(Number(versionNumberString))
+  })
+
+  const previousVersion = tags.length > 0 ? tags[tags.length - 1] : 0
```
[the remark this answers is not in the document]
ckline-tryptic: Yes, the ref's dont have any timestamps or other ordering information I could find, and, it does appear to be in order.

client/src/push-content.ts
rnathuji: Hmmm...I don't see how that impacts my question here. Stated differently, I think it's possible given this code for a user to expect a tag but for one not to be applied because `_getTag` can return `undefined`. I think either the return signature of `_getTag` should be modified to not allow returning `undefined`, or an explicit check needs to happen after L98 to throw an error if the user clicks wanting a tag but (for whatever reason) `getTag` returns `undefined` and this branch isn't taken.

client/src/push-content.ts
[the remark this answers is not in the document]
rnathuji: Right, but it's at least possible for the tags to be out of order, correct? Since there's an underlying assumption here that these tags will be numbers anyways, would it be safer to take the largest number in the array instead of the last entry just in case?

client/src/push-content.ts
rnathuji: Oh, actually a follow up question regarding workflow: If a user wants to commit + tag on a push, doesn't the tagging need to happen after creating the commit, or am I misunderstanding something here?

client/src/push-content.ts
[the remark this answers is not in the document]
rnathuji: 🤔 - I don't know that it's in the spirit of the CI status check to use these ignores as placeholders for future tests outside of a PR, though perhaps that's something to discuss with the team in a 🅿️ . In any case, I'd be happy to pair if that helps as far as getting test coverage into this PR itself.

client/src/push-content.ts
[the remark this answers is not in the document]
ckline-tryptic: makes sense

client/src/push-content.ts
ckline-tryptic: No misunderstanding.

package.json
```diff
@@ -104,7 +110,7 @@
     "viewsWelcome": [
       {
         "view": "openstax-controls",
-        "contents": "[Open ToC Editor](command:openstax.showTocEditor)\n[Open Image Upload](command:openstax.showImageManager)\n[Push Content](command:openstax.pushContent)"
+        "contents": "[Open ToC Editor](command:openstax.showTocEditor)\n[Open Image Upload](command:openstax.showImageUpload)\n[Push Content](command:openstax.pushContent)\n[Tag Content](command:openstax.tagContent)"
```
rnathuji, line 113: I think something went wrong in a merge along the way as this should stay `showImageManager` (the button is otherwise broken on this branch).

rnathuji (commented):
Overall the updated PR looks good, there's just one minor [outstanding comment](https://github.com/openstax/poet/pull/49#discussion_r616751460) from earlier and [an issue](https://github.com/openstax/poet/pull/49#discussion_r625107398) I noticed on this pass that probably happened accidentally during a merge.

rnathuji (approved):

PR #58, opened 2021-04-28
Temporarily include title tag in module template

tomjw64:


client/src/panel-toc-editor.ts
```diff
@@ -71,15 +71,16 @@ const requestBundleModules = async (client: LanguageClient, args: BundleModulesA
...
-      <metadata xmlns:md="http://cnx.rice.edu/mdml">
-        <md:title>New Module</md:title>
-        <md:content-id>${newModuleId}</md:content-id>
-        <md:uuid>${uuidv4()}</md:uuid>
-      </metadata>
-      <content>
-      </content>
-    </document>`.trim()
+<document xmlns="http://cnx.rice.edu/cnxml">
+  <title>${newModuleId}</title>
```
rnathuji, line 75: 🤔 - Why is the `<title>` the generated `<md:content-id>` vs matching `<md:title>`?

tomjw64: Oh my, I forgot that we are removing `md:title` and not `title`. I thought this was just temporary and we'd want the value to be as unique as possible since the slug is generating from it. But it is most definitely not temporary.

rnathuji (approved):

PR #60, opened 2021-04-30
Detect and handle bundle directory deletions

rnathuji:
This PR attempts to handle an uncommon case reasonably, without accounting for all corner cases (e.g. a directory is accidentally moved and then moved back, but the latter doesn't necessarily seem to trigger a create event for each individual file in all situations) to keep things simple for now. Specifically, per comments in the [parent issue](https://github.com/openstax/cnx/issues/1510), it seems the current file watcher implementation in VScode may not always send all file-level changes when a directory is deleted or renamed. The first scenario in the original description in the bug issue was an instance of a side effect of this limitation. In this PR I tried to handle some limited cases to ensure our bundle model (and consequently validations) at least get triggered to update. In practice I'd expect these situations would only arise if:

* Modules - A user deletes a module directory after creating (this seems like a pretty reasonable thing we want to make sure is covered and the main impetus for this PR) or accidentally renames / drags and drops it somewhere
* Collections - A user accidentally renames or drags and drops this somewhere
* Media - A user accidentally renames or drags and drops this somewhere

### DIFF
--- a/server/src/book-bundle.ts
+++ b/server/src/book-bundle.ts
@@ -630,7 +630,10 @@ export class BookBundle {
       return true
     }
 
-    return (deletedPath.includes(this.moduleDirectory()) && !deletedPath.endsWith('index.cnxml'))
+    const indexOfLastSep = deletedPath.lastIndexOf(FS_SEP)
+    const maybeModuleId = deletedPath.substring(indexOfLastSep + 1)
+    return ((deletedPath === this.moduleDirectory()) ||
+            (deletedPath.includes(this.moduleDirectory()) && this.moduleExists(maybeModuleId)))
   }
 
   processDirectoryDeletion(change: FileEvent): void {
--- a/server/src/book-bundle.ts
+++ b/server/src/book-bundle.ts
@@ -618,6 +618,45 @@ export class BookBundle {
     }[item.type][change.type].bind(this)
     func(item.key)
   }
+
+  isDirectoryDeletion(change: FileEvent): boolean {
+    if (change.type !== FileChangeType.Deleted) {
+      return false
+    }
+    const deletedPath = URI.parse(change.uri).fsPath
+
+    // This assumes both collections and media dirs are flat
+    if ((deletedPath === this.collectionDirectory()) || (deletedPath === this.mediaDirectory())) {
+      return true
+    }
+
+    return (deletedPath.includes(this.moduleDirectory()) && !deletedPath.endsWith('index.cnxml'))
+  }
+
+  processDirectoryDeletion(change: FileEvent): void {
+    const deletedPath = URI.parse(change.uri).fsPath
+
+    if (deletedPath === this.collectionDirectory()) {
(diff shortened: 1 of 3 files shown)


server/src/book-bundle.ts
```diff
@@ -618,6 +618,45 @@ export class BookBundle {
...
+      return false
+    }
+    const deletedPath = URI.parse(change.uri).fsPath
+
+    // This assumes both collections and media dirs are flat
+    if ((deletedPath === this.collectionDirectory()) || (deletedPath === this.mediaDirectory())) {
+      return true
+    }
+
+    return (deletedPath.includes(this.moduleDirectory()) && !deletedPath.endsWith('index.cnxml'))
```
tomjw64, line 633: If someone mistakenly drops a non-`index.cnxml` file in a module directory and then deletes it, will this cause the model to try to drop a module that doesn't exist (which I don't believe will error, but it will bust the cache)?

i.e. should we include the following in our tests?
```
assert(!bundle.isDirectoryDeletion({ type: FileChangeType.Deleted, uri: '/bundle/modules/m00001/random.txt' }))
```
I see you explicitly mentioned this is NOT a catch-all solution, though. This may be one of those corner cases we don't care about, since invalidating the cache for a single module isn't really a huge deal.

server/src/server.ts
```diff
@@ -126,7 +126,16 @@ connection.onDidChangeWatchedFiles(({ changes }) => {
         return
       }
       const [bundleChanged, bundleValidator] = expect(workspaceBookBundles.get(workspaceChanged.uri), 'already returned if key missing')
-      bundleChanged.processChange(change)
+      if (bundleChanged.isDirectoryDeletion(change)) {
```
tomjw64, line 129: Thoughts on moving this branching into the `processChange` function as a guard, so we don't have to meet "the change uri is not a directory" as an invariant for calling it and getting correct behavior? maybe it makes no difference if this is the only call site...

server/src/book-bundle.ts
rnathuji: Hmmm...good point. As you suspect, currently the test case would fail, `onModuleDeleted(moduleId)` would get called with `foo.txt` not impacting the list of modules but busting the cache. Maybe instead of testing `!deletedPath.endsWith('index.cnxml'))` I can test that end of the path is a valid module ID to address the test case + be safer in general.

server/src/server.ts
rnathuji: Sure, we can do that.

server/src/book-bundle.ts
rnathuji: [Done](https://github.com/openstax/poet/pull/60/commits/7b5a17630eaa6e821564f8034418bf0ff945da38)

server/src/server.ts
rnathuji: [Done](https://github.com/openstax/poet/pull/60/commits/8855d89caf6bd40fddbd998e917271bbba6e0dbf)

tomjw64 (approved):

PR #63, opened 2021-04-30
Ignore empty module directories in book bundle model

rnathuji:


server/src/book-bundle.ts
```diff
@@ -314,7 +314,8 @@ export class BookBundle {
       }
     }
     const loadModules = async (bundle: BookBundle, map: Map<string, ModuleInfo>): Promise<void> => {
-      const foundModules = await fs.promises.readdir(bundle.moduleDirectory())
+      const foundPossibleModules = await fs.promises.readdir(bundle.moduleDirectory())
+      const foundModules = foundPossibleModules.filter((moduleId) => fs.existsSync(path.join(bundle.moduleDirectory(), moduleId, 'index.cnxml')))
```
tomjw64, line 318: Can we do this asynchronously and in parallel for performance's sake (since startup time for the extension can already be fairly significant in a worst case scenario)? We have a function in `utils` called `fileExistsAt` that uses the async stat function to tell if a file exists (since `fs.exists` is deprecated and `fs.promises.exists` doesn't itself exist).

rnathuji: [Yep](https://github.com/openstax/poet/pull/63/commits/50501f3843f65a13c8f2d89f9a38281c11f05962), but LMK if there's a better / cleaner idiom than that.

tomjw64 (approved):

PR #65, opened 2021-05-10
Bn/cnx#1491/catch xml errors

c0ldcalamari:


server/src/bundle-validation.ts
```diff
@@ -81,6 +82,16 @@ export class BundleValidationQueue {
   }
 }
 
+export const collectionDiagnostic = async (): Promise<Diagnostic[] | null> => {
```
rnathuji, line 85: Minor: Could the return type here just be `Promise<Diagnostic[]>` since I believe it's definitely going to return a non-null value?

server/src/server-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+    const bundleAndValidator = workspaceBookBundles.get(request.workspaceUri)
+    if (bundleAndValidator == null) { return null }
+    const bundle = bundleAndValidator[0]
+    const promises = bundle.collectionItems().map(async (collection: BundleItem): Promise<TocTreeCollection[]> => {
+      try {
+        const tree = expect(await bundle.collectionTree(collection.key), 'collection must exist').inner
+        return [tree]
+      } catch {
+        const uri = expect(bundle.bundleItemToUri(collection), 'No root path to generate diagnostic')
+        const diagnostics = expect(await collectionDiagnostic(), 'No diagnostic to generate')
```
rnathuji, line 29: This expect can be removed if the return type of `collectionDiagnostics()` is modified per separate comment.

server/src/server-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+        })
+        return []
+      }
+    })
+    const trees = (await Promise.all(promises)).flat()
+    return trees
+  }
+}
+
+export function bundleOrphanedModulesHandler(workspaceBookBundles: Map<string, [BookBundle, BundleValidationQueue]>): (request: BundleOrphanedModulesArgs) => Promise<BundleOrphanedModulesResponse> {
```
rnathuji, line 42: Yep, it looks like codecov is dinging you because of these handlers being refactored. Per discussion during standup, you could alternatively leave these two where they were and create a separate issue to pull them out + add tests.

server/src/test/server.test.ts
```diff
@@ -16,15 +15,127 @@ import {
...
+    const invalidTitleUri = 'file:///bundle/collections/invalidtitle.xml'
+    const workspaceBookBundles: Map<string, [BookBundle, BundleValidationQueue]> = new Map()
+    workspaceBookBundles.set('file:///bundle', [bundle, validationQueue])
+    const handler = bundleTreesHandler(workspaceBookBundles, connection as any)
+    const request = { workspaceUri: 'file:///bundle' }
+    const bundleTreesResponse = await handler(request)
+    const diagnostic = await collectionDiagnostic()
+    assert(bundleTreesResponse)
+    assert(connection.sendDiagnostics.calledTwice)
+    assert(connection.sendDiagnostics.withArgs({
```
rnathuji, line 97: I don't think this assertion is doing quite what you want (e.g. if you make `onFirstCall()` `onSecondCall()` here, the assertion will still pass). Perhaps these match your intent in these two test cases?:

```javascript
    assert(connection.sendDiagnostics.getCall(0).calledWithExactly({
      diagnostics: diagnostic,
      uri: invalidSlugUri
    }))
    assert(connection.sendDiagnostics.getCall(1).calledWithExactly({
      diagnostics: diagnostic,
      uri: invalidTitleUri
    }))
```

rnathuji (commented):
Looks good overall, just a few minor comments.

rnathuji (approved):

c0ldcalamari:
Test for Otto:

# Test 1
## Given:
The TOC tree / collection.xml in POET

## When: 
The collection.xml throws a parsing error/ title or slug becomes unparsable - 

```
<<</<</<<<<<md:slug>business-law-i-essentials</md:slug>
```

## Then: 
- The user is notified with diagnostics in the "problems" panel
- VScode will NOT throw a hard error 

(The below should not happen)
Screen Shot 2021-04-16 at 10 47 02 AM


---

# Test 2

## Given:
A bundle/repo with more than one collection in POET

## When: 
The collection.xml throws a parsing error/ title or slug becomes unparseable - 

```
<<</<</<<<<<md:slug>business-law-i-essentials</md:slug>
```

## Then:
- The user is notified with diagnostics in the "problems" panel
- VScode will NOT throw a hard error 
- All collections where no errors return will return. 

Example: If in the university physics bundle, if university-physics-1 has an error, but university-physics-2 & university-physics-3. Then in the TOC editor, you will still receive the collections where no error has occurred.

PR #66, opened 2021-05-12
Limit directories to commit / push with git add

rnathuji:
This PR modifies the `Push` button to only stage / commit files in a specific set of directories. However, it doesn't seem like the exposed `commit()` method provides a way to do this scoping, so instead it uses the private `add` method which isn't ideal (we may want to create an issue similar to [cnx 1519](https://github.com/openstax/cnx/issues/1519) to upstream exposing this in the public API). A benefit, however, is we can explicitly check for staged differences to determine if there are changes to push prior to prompting for a commit message versus checking the `stdout` of errors after attempting to commit.

The PR doesn't make any changes to the tagging flow as that currently uses `diffWithHEAD()` to detect un-pushed changes which should continue working as desired.

### DIFF
--- a/client/src/push-content.ts
+++ b/client/src/push-content.ts
@@ -127,6 +127,10 @@ export const _pushContent = (
   }
 
   await gitAddCmd(validContentUris)
+  if ((await repo.diff(true)).length === 0) {
+    void errorReporter('No changes to push.')
+    return
+  }
 
   let commitSucceeded = false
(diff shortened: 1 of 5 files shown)


rnathuji:
Closing this PR for the time being per [this comment](https://github.com/openstax/cnx/issues/1538#issuecomment-840695219)

PR #73, opened 2021-06-02
Fix issues related to running client tests in local environments

rnathuji:
This PR addresses a few failures that were observed when running client tests locally (though all pass on CI) due to the following factors:

* Running tests on macOS
* Running tests with a newer version of VSCode (relative to `1.54.1` which the CI [currently runs with](https://github.com/openstax/poet/blob/471c372711b1961a38e121720e00afc98071cb11/client/src/test/runTest.ts#L7))

It also goes ahead and updates the version of vscode used for tests to `1.56.2`.


m1yag1:
We need to create a card that has details regarding improvements that can be made. I think we need to spend a little time creating a waiting function/coroutine that can check for a particular value to return true and timeout if it does not within some timeout limit.

PR #74, opened 2021-06-02
changee dialoge into modal

c0ldcalamari:


rnathuji (approved):

c0ldcalamari:
For Otto:
Multiple Dialogues should now longer appear, only one dialogue should prompt the user to make a choice before they can continue.

PR #77, opened 2021-06-14
make upload image functionality invisible for now

therealmarv:
for details see https://github.com/openstax/cnx/issues/1579


rnathuji (approved):
Just a passing thought: Depending on whether CMs know about using the command palette or not, is it worth removing the image upload command for the time being as well 🤷‍♂️?

therealmarv:
I will remove that too! Good catch...

rnathuji (approved):

PR #80, opened 2021-06-17
Address preview styling issues introduced by XSL and generic styles

rnathuji:


client/src/webview-js/cnxml-preview/content-common.less
```diff
@@ -114,7 +114,7 @@ figure:not([data-orient="vertical"]) {
   }
 
   > figcaption {
-    display: table-caption;
+    // display: table-caption;
```
m1yag1, line 117: Do you mean to have this comment here? I tend to shy away from keeping commented code around but of course, there are certain conditions when we can make exceptions.

m1yag1 (approved):

client/src/webview-js/cnxml-preview/content-common.less
rnathuji: Yeah -- I left it as a comment thinking just in case the file is copied again [from REX](https://github.com/openstax/rex-web/tree/master/generic-styles/) it might be somewhat clearer it was a change we made vs a diff introduced by changes in that repo 🤷‍♂️ (or at least trigger someone to do a `git blame` on the line in question).

m1yag1: It's probably fine then.

PR #86, opened 2021-06-29
Add git history as dependency and include button

rnathuji:


rnathuji:
The extension seems to run fine in my local vscode, but I noticed on gitpod sometimes the history window comes up empty:



I'm dropping this observation in the issue for now, and we can see if it's reproduced by others. I've also let @ne-newton know in case he wants to try out the package from the PR on his system.

ckline-tryptic (approved):

PR #92, opened 2021-07-12
Make POET repo a valid book again

rnathuji:


.vscode/settings.json
```diff
@@ -3,9 +3,6 @@
     "files.associations": {
         "*.cnxml": "xml"
     },
-    "xml.catalogs": [
```
m1yag1, line 6: This was needed before to fix the XML error that would occur when first loading POET. Is this not needed anymore?

rnathuji: Nope -- starting with [this PR](https://github.com/openstax/poet/pull/85) we actually tell the XML server about the catalog location through the API. That allows us to remove it here and also avoids a race condition that can happen in certain scenarios.

m1yag1: I see. I was mostly being curious ;)

m1yag1 (approved):

PR #95, opened 2021-08-02
Image autocomplete

ckline-tryptic:


rnathuji (commented):
I have a somewhat general question on this PR which is what the expected behavior should look like? I tried both gitpod and local VS code and seemingly wasn't able to get autocomplete hints (though local VS code I think is related to the filesystem related comments below) 🤷‍♂️ . It may well be human error / setup issues on my side, though.

philschatz (changes requested):

philschatz (approved):
Looks great! Want to rebase+squash all the fixup commits?

I think there's a way for git to automatically combine them with the original commit

PR #98, opened 2021-08-12
Use Jest for server tests

philschatz:
[Jest](https://jestjs.io/) significantly shrinks the dev + test feedback loop:

1. code does not have to be rebuilt because [ts-jest](https://github.com/kulshekhar/ts-jest) transpiles on the fly
    - no more `npm run build && npm run test:server` & wait
1. `jest --watch` reruns tests when the code changes
1. `jest --watch` smartly re-runs a subset of tests
    - Jest records which code files were used when running each test

Try it out by [opening this PR in gitpod](https://gitpod.io/#https://github.com/openstax/poet/pull/98), running `npm run test:server:watch`, and editing a file in `server/src/`

## Notes

It would be nice to use Jest in the other tests but that may be a bit more challenging:

- cypress uses mocha exclusively but those are browser tests and require the slow `npm run build` anyway
- the client tests that cannot use a mocked `vscode` package need to run inside vscode and likely require the slow `npm run build` as well
- the client tests that can use a mocked `vscode` package could be pulled out and Jest-ified

Converting from Mocha to Jest is mostly just a few replacements but there are 2 tricky bits:

1. Both mocha and jest define global variables which conflict and permeate the whole tsconfig.json scope. Multiple tsconfig.json files might keep the scope separated
1. ts-jest 27 might be broken. I spent way too much time chasing random phantom lstat errors. But now they appear to be gone

PS: I haven't used it much yet but [vscode-jest extension](https://github.com/jest-community/vscode-jest) offers a bit more editor integration.


rnathuji (approved):
Cool!

ckline-tryptic (approved):

PR #104, opened 2021-09-09
Fix Langauge Server errors that arise during testing

philschatz:
- run fix-document-ids only on files that actually exist
- send `Diagnostics.source = 'cnxml'` so pushContent can prevent users from pushing errors


therealmarv (approved):

therealmarv:
codecov down?

ckline-tryptic (approved):

PR #105, opened 2021-09-09
upgrade codecov action to v2

philschatz:
Codecov is not reporting when commits are pushed so I thought this might be why (looks like it's not the reason).

Also, codecov has deprecated v1 of their action: https://github.com/marketplace/actions/codecov


m1yag1 (approved):

m1yag1:
@philschatz this is approved.

PR #116, opened 2021-11-15
:bug: fix CI failure on `main` branch

philschatz:
The `main` branch has a test failure in CI (GitHub Actions). I was not able to reproduce locally so this PR should help in diagnosing the error


philschatz:
The error seems to have been transient. If it appears again we could reopen this PR

PR #119, opened 2021-12-07
validate first Page in chapter has intro class

philschatz:
# To-Do

- verify it works in books with units
- verify it doesn't give false positives for any of our books

### DIFF
--- a/server/src/model/page.ts
+++ b/server/src/model/page.ts
@@ -1,8 +1,9 @@
-import I from 'immutable'
+import I, { hasIn } from 'immutable'
 import * as Quarx from 'quarx'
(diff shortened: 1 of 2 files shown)


server/src/model/page.ts
```diff
@@ -170,10 +170,12 @@ export class PageNode extends Fileish {
       }
     })))
 
+    const introClasses = ['introduction', 'unit-opener']
```
philschatz, line 173: I think a unit-opener can only occur inside a unit and a chapter introduction can only exist in a chapter. Lemme ask why the term "unit-opener" was used instead of "introduction" inside a Unit

PR #120, opened 2021-12-08
Reload workspace settings on activation

TylerZeroMaster:
This is one part of a quick solution to our GitPod problems. There might be a more permanent solution available; however, I have not found one. 

With this change, activating POET will cause vscode to reload the settings.json file. This should fix file associations not working correctly.

### DIFF
--- a/client/src/test/suite/extension.test.ts
+++ b/client/src/test/suite/extension.test.ts
@@ -5,7 +5,7 @@ import vscode from 'vscode'
 import SinonRoot from 'sinon'
 import { GitErrorCodes, Repository, CommitOptions, RepositoryState, Branch, RefType } from '../../git-api/git.d'
 import 'source-map-support/register'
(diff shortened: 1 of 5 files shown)


client/src/test/suite/extension.test.ts
```diff
@@ -174,6 +174,7 @@ suite('Extension Test Suite', function (this: Suite) {
   })
   test('show cnxml preview with no file open', async () => {
     assert.strictEqual(vscode.window.activeTextEditor, undefined)
+    await sleep(500) // FIXME: Make me go away (see https://github.com/openstax/cnx/issues/1569)
```
TylerZeroMaster, line 177: I added this only because it seemed like client tests (`npm run test:client`) were consistently failing due to a race condition.

philschatz (approved):

PR #124, opened 2022-01-25
:fire: remove git-history-plus

philschatz:
because that extension does not exist in the vscode marketplace.

This makes it possible for people to install the extension from the VSCode marketplace


m1yag1:
@ckline-tryptic suggested this plugin that is available on the marketplace. https://marketplace.visualstudio.com/items?itemName=mhutchie.git-graph

m1yag1:
**Reminder** When we remove this extension we need to remove the button.

TylerZeroMaster:
It also might be worth noting that [donjayamanne.githistory](https://marketplace.visualstudio.com/items?itemName=donjayamanne.githistory) is on both marketplaces and has the exact same commands as git history plus. This could be an easy drop-in replacement if we want to go that route.

ne-newton:
I have confirmed with CMs: none of them have used this, so it can safely be removed, along with the button.

sparksam (approved):

ckline-tryptic (approved):

therealmarv (approved):

PR #125, opened 2022-02-08
:bug: ignore absolute iframes

philschatz:
refs openstax/ce#1813

### DIFF
--- a/server/src/model/page.spec.ts
+++ b/server/src/model/page.spec.ts
@@ -85,6 +85,23 @@ describe('Page validations', () => {
     image.load('somebits')
     expect(page.validationErrors.errors.size).toBe(0)
   })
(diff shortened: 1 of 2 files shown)


TylerZeroMaster (approved):
Looks good! I guess it would be questionable to have POET *automatically* send hundreds of GET or HEAD requests to validate that external content exists.

philschatz:
@TylerZeroMaster Yep, but.... the [CLI version](https://github.com/openstax/poet/tree/main/server#lint-a-book-and-find-broken-redirects) of the language server does [exactly that](https://github.com/openstax/poet/blob/main/server/src/model/_cli.ts#L105)! With it we also found links that had [2+ redirects](https://github.com/openstax/poet/blob/main/server/src/model/_cli.ts#L127) and reduced them down to like 1

PR #136, opened 2022-04-05
Fix CI and check if xsd directory exists before removing it

philschatz:
Refs openstax/ce#1872


client/src/extension.ts
```diff
@@ -44,24 +44,45 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
...
+  /* The following istanbul comments are a hack because the coverage seems to misrepresent them as uncovered */
+  /* istanbul ignore next */
   if (workspaceRoot !== null && fs.existsSync(path.join(workspaceRoot.fsPath, 'META-INF/books.xml'))) {
     await populateXsdSchemaFiles(resourceRootDir)
   }
 
+  /* istanbul ignore next */
+  const extExports = doRest(client)
+  /* istanbul ignore next */
   await client.onReady()
```
TylerZeroMaster, line 56: I tested this version in GitPod. The .xsd folder is created when it does not exist and the files are populated, but the client does not start (i.e. no ToC tree, buttons do nothing). I looked at my browser's console and saw this error: `Activating extension 'openstax.editor' failed: Language client is not ready yet.`

Previously, this `await client.onReady()` occurred before `client.onRequest(...)` and `client.onNotification(...)`. I am not sure if rearranging these calls would fix the issue, or if this issue might somehow relate the to code coverage problems you dealt with, but it seems plausible.

TylerZeroMaster (changes requested):

client/src/extension.ts
philschatz: Hmm, thanks for checking in GitPod. I rearranged the order and will try it in GitPod

philschatz: Loading the extension in GitPod works for me. Could you try installing this new version?

TylerZeroMaster (approved):

PR #137, opened 2022-04-14
fix webview disposal problems

philschatz:
When a user opens and then closes a Preview panel they may get an error because POET was trying to communicate with a disposed webview.

This checks if the webview has been disposed before trying to change the webview.

### DIFF
--- a/client/specs/extension.spec.ts
+++ b/client/specs/extension.spec.ts
@@ -106,7 +106,7 @@ describe('Extension', () => {
         const pm = extensions[OpenstaxCommand.SHOW_CNXML_PREVIEW] as PanelManager<CnxmlPreviewPanel>
         const panel1 = pm.revealOrNew()
         const panel = (panel1 as any).panel as WebviewPanel // `.panel` is a protected field
-        expect(panel.webview.html.includes('No resource available to preview')).toBeTruthy()
+        expect(panel.webview.html).toEqual(expect.stringContaining('No resource available to preview'))
       })
 
       it('show cnxml preview with a file open', async () => {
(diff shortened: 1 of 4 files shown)


m1yag1 (approved):

TylerZeroMaster:
Unfortunately, I am still seeing this error after applying these fixes.
Screen Shot 2022-04-15 at 11 14 24 AM

TylerZeroMaster:
I think I have managed to gain a greater understanding of the problem. The preview panel has an event listener for `onDidChangeActiveTextEditor`. When you close the preview while it is focused, this event fires twice in rapid succession (once for the preview panel closing and once for the new editor becoming focused). If you focus a different editor before closing the preview, you do not get an error and you only get 1 `onDidChangeActiveTextEditor` event.

Long story short, this fixes it (although, it looks like [these](https://github.com/openstax/poet/pull/137/files#diff-ae4d88dbbedd30566f6e35512bfb0213c2d30471fab3c6ece501014e0998c1beR154-R157) lines are supposed to accomplish the same thing):
```ts
    this.registerDisposable(vscode.window.onDidChangeActiveTextEditor((editor) => {
      if (editor?.document.uri.fsPath === this.resourceBinding?.fsPath) {
        return
      }
      void ensureCatchPromise(this.tryRebindToActiveResource(false))
    }))
```

Event loop 👻

philschatz:
I found something related too but it requires adding an `await sleep(1)` to the top of `tryRebindToActiveResource`.

You mentioned yesterday that vscode's Markdown Preview avoids this problem by delaying the execution of the update code, giving a chance for the view's `dispose` function to run.

I'll try your `if (editor?.document.uri.fsPath === this.resourceBinding?.fsPath)` and if it doesn't work then a sleep(1) or delayed timer would seem like an okay fallback.

TylerZeroMaster (approved):

PR #138, opened 2022-04-19
1880 Fix validate content

TylerZeroMaster:


philschatz (approved):
Yay! And no more detecting which documents are open!

philschatz (approved):

PR #139, opened 2022-04-20
Autogenerate IDs quickly

philschatz:
refs openstax/ce#1882

Implemented as a Hint. Hint shows up just barely next to the `<` and does not show up in the Problems area:

Screenshot from 2022-05-05 17-05-01

This does 3 things:

- add a button to `[Auto-ID Content]` instead of generating IDs when pushing content
- only update files that the model already knows are missing IDs
- add a warning when elements are missing an ID attribute
    - could be an informational message instead of a warning

## TODO

- [x] Figure out why code coverage is being misreported for `extension.ts` or at least fix it for this PR

# Alternatives

Could have been "Informational". That would be a blue line and an entry in the Problems area like so:

image


philschatz:
Cross-posting from [Slack Conversation](https://openstax.slack.com/archives/C0LA54Q5C/p1650556489569729):

The code coverage mangling is at the bottom of client/src/extension.ts .
This happened to me before [last week](https://github.com/openstax/poet/blame/main/client/src/extension.ts) (see `Fix CI` lines) and tried the following techniques just to get an idea of what was going on:

I tried adding `/* ignore next */` lines everywhere coverage errors were reported but that did not remove the error
It seemed that the return types were causing coverage to mangle... So I created these pretty useless [little functions](https://github.com/openstax/poet/commit/24ce6f75afe1f3838bab62007c30e023a37bc4a8#diff-7b7aa9c8b1944a19471cbe4172196463223080d88b4433c79933f3a47f8ca0f3R76) and [this](https://github.com/openstax/poet/commit/24ce6f75afe1f3838bab62007c30e023a37bc4a8#diff-7b7aa9c8b1944a19471cbe4172196463223080d88b4433c79933f3a47f8ca0f3R112), and then ended up with lines like [this](https://github.com/openstax/poet/commit/24ce6f75afe1f3838bab62007c30e023a37bc4a8#diff-7b7aa9c8b1944a19471cbe4172196463223080d88b4433c79933f3a47f8ca0f3R55-R58)

Some ideas I was going to pursue:

This is the only file that has coverage problems. This file runs in the really slow start-up-vscode-instead-of-mocking-it. In order to run in vscode the file is transpiled into JS and then instrumented with coverage-recording in the [scripts/pretest.sh](https://github.com/openstax/poet/blob/main/scripts/pretest.bash#L15-L16) file. Maybe there's a transpile problem, instrumentation problem, or maybe vscode injects some JS at the top of the file before running?
Since the instrumented extension.js file runs in VSCode, there's a little bit of [custom JS](https://github.com/openstax/poet/blob/main/client/src/test/suite/index.ts#L41-L50) that reports the code coverage back out of the tests once they complete here
Finally, this (really slow) method of running unit tests inside vscode starts [here](https://github.com/openstax/poet/blob/main/client/src/test/runTest.ts) and was taken a couple years ago from vscode's [documentation](https://code.visualstudio.com/api/working-with-extensions/testing-extension).


Also, the coverage from jest & vscode's method are merged at the end using `nyc merge` [here in package.json](https://github.com/openstax/poet/blob/main/package.json#L43)

At some point it's less effort to port the remaining tests here into jest.

(Tyler converted about half the tests since this was written so porting the remaining tests might be easy)

philschatz:
The code coverage error seems to [only be fixable](https://github.com/openstax/poet/pull/142#issuecomment-1117402292) by porting all the vscode tests to jest

I'll try moving the "port tests to Jest" into a separate PR

philschatz:
Currently waiting on #143

PR #140, opened 2022-04-22
Port the remaining push button client tests

TylerZeroMaster:
We talked about trying to eliminate the tests in `client/src/test/suite/extension.test.ts`. They run in a vscode process and tend to be problematic for a variety of reasons. I thought porting the remaining push button tests would help.


philschatz (approved):
Awesome!

PR #141, opened 2022-05-02
Attempt to fix coverage issues. 

sparksam:
`npm run test:unit`  (`jest --coverage --coverageDirectory ./.nyc_output/`) pass the coverage as the `extension.js`  gets a 100% coverage but the `npm run pretest:client`  (`/home/runner/work/poet/poet/node_modules/.bin/nyc instrument --exclude 'client/out/client/src/test/**/*' --exclude-node-modules --compact=false --source-map --in-place ./client/out/ ./client/out/`) instruments the client code and the following test and `nyc merge` breaks the `extension.js` coverage

I found that sometimes `nyc coverage`  show [incorrect line numbers](https://stackoverflow.com/questions/44120502/nyc-coverage-shows-incorrect-line-numbers) due to instrumentation and some use `babel-plugin-istanbul` for instrumentation. Which I realized is a dependency of jest  I tried to follow [this](https://github.com/istanbuljs/babel-plugin-istanbul)  and run the tests and skip this [line](https://github.com/openstax/poet/blob/ff02277609217030ff78b96b7871474c401ea7bc/scripts/pretest.bash#L16) in the pretest.bash but no success so far. 

Current state of things: I can't have `nyc` find the `babel-register` module. @philschatz Please can you have a look?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,20 @@
+{
+  "env": {
+    "test": {
+      "plugins": [
+        [
+          "istanbul",
+          {
+            "exclude": [
+              "node_modules/",
+              "client/out/client/src/test/**/*",
+            ]
+          },
+          {
+            "useInlineSourceMaps": false
+          }
+        ]
+      ]
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


TylerZeroMaster:
Adding `--require @babel/register` (`NODE_ENV=test $(npm bin)/nyc --require @babel/register instrument`) fixes the module not being found. Unfortunately, that results in a new error `'Did not find coverage data on global.__coverage__ . Failing'` originating from `client/src/test/suite/index.ts`.

Out of curiosity, I also tried running `npx nyc --require @babel/register npm run test:client` and the result of that command was the same.

philschatz:
Closing in favor of #143

PR #142, opened 2022-05-02
Transpile using babel only for more precise code coverage reporting

philschatz:
Jest tests use babel, Webpack uses babel, and Cypress uses babel. The only part that does not use babel is the client tests that run in vscode.

This PR:

1. changes it so that the client is transpiled using babel too
2. code coverage is injected via the `babel-plugin-istanbul` plugin when `NODE_ENV=test` is set

# TODO

- [x] Remove configs that are no longer used.
    - webpack config is configured to use several plugins and configures jsx to use Preact
 - [x] Verify that TypeScript errors will cause the automated tests to fail (babel may not perform type checking)
     - Webpack catches it. Not ideal, but good enough!
 - [x] make usre launching via VSCode launch.json still works


ckline-tryptic (approved):

philschatz:
Closing in favor of moving this work into #143 because we need to port the tests to Jest (for codecov) before replacing tsc with babel

PR #143, opened 2022-05-04
Replace slow vscode tests with Jest

philschatz:
This has several benefits:

- Tests can run in watch mode
- Code coverage is reported correctly
- Compiling the code is no longer necessary to run tests

## TODO

These are optional but the work was already done in another PR:

- [ ] completely remove `tsc` and always use babel to transpile TypeScript into JavaScript (we already use it for the tests, the webview files, and the webpacked extension)
- [ ] use babel-istanbul plugin to inject code coverage for the webview tests
- [ ] rearrange the test scripts so we run webpack right before the cypress tests instead of at the beginning


philschatz:
Currently waiting on openstax/poet#143 to be merged. This issue is waiting for the code coverage to be fixed (by removing one testing system in POET and bringing it down to 2)

sparksam (approved):
Looks good to me.

PR #151, opened 2022-08-05
Add MathML validator class that encapsulates math validation

TylerZeroMaster:
https://user-images.githubusercontent.com/33585550/183145965-189b73b3-f3fe-44f5-beb1-5687d70b60f4.mov

Somewhat clunky: page.ts uses ensureLoaded to check if fields are
loaded; however, that function is protected.

I still like the idea of encapsulating math validation in an external
class to keep things neat and decoupled, but it needs more work.

### DIFF
--- a/server/src/model/external-validator.ts
+++ b/server/src/model/external-validator.ts
@@ -1,7 +1,0 @@
-import { ValidationCheck } from './fileish'
-import { PageNode } from './page'
-
-export interface ExternalValidator {
-  parseXML: (doc: Document) => void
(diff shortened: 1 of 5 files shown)


philschatz:
Are these validations that cannot be (easily?) done in XSD? Something like:

```xml
  <xsd:simpleType name="positiveDecimal">
    <xsd:restriction base="xsd:decimal">
      <xsd:minInclusive value="0"/>
    </xsd:restriction>
  </xsd:simpleType>
```

PR #155, opened 2023-01-31
In gitpod, port 6009 was not working for debugging, use 16009 instead

TylerZeroMaster:
Adding a 1 in front seemed to fix the problem. It was probably some kind of port conflict. After this change, we do not need a gitpod-specific launch configuration either.


sparksam:
@TylerZeroMaster This can be merged, I think.

sparksam (approved):
Looks good to me.

PR #156, opened 2023-02-01
Validate resource filename in Resource; Update Diagnostics all the time

philschatz:
This puts the resource-filename-validation code in the resource, ensures that all Resources are loaded into the model, and sends an updated Diagnostics element whenever the model changes

We discussed reducing the overhead of computing a Diagnostics message on every character change but figured we should try it on a real book first


sparksam (approved):
Looks good to me.

PR #157, opened 2023-03-09
Set pull.ff if pull.rebase and pull.ff are unset

TylerZeroMaster:
- [ ] openstax/ce#2030

Prevent fatal error on content push caused by the strategy for handling divergent branches being unset. I tried to be non-intrusive by only setting our default value when the other options have not been set.

I added steps to recreate the problem to the issue. You can try it the same way, using a local remote, with POET.


ckline-tryptic (approved):

omehes:
verified with editor-0.0.0-dev-d4f9a43.vsix
ok

PR #158, opened 2023-03-15
Make content-id optional in collxml schema

TylerZeroMaster:
Since this element is not used anymore and the ToC editor omits it when it rewrites the ToC, it should be optional until it can be completely phased out.

See related PR:
- openstax/cnxml#61


philschatz (approved):

omehes:
testing - editor-0.0.0-dev-d4f9a43.vsix
osbooks-otto-books

verified by removing all content-id, readding, rearranging modules via TOC editor, changing module titles, etc
OK

PR #160, opened 2023-05-02
2048 - Generate readme

TylerZeroMaster:
fixes openstax/ce#2048

~The readme generator is mostly the same as [here](https://github.com/openstax/ce-scripts/blob/master/git-repo-prep/index.ts). The main difference is using `vscode.workspace.fs` instead of `node:fs` for reading/writing files.~ 

I decided to remove `tagContent` in the process since I had already repurposed the `Build in CORGI` button.

## Demo:

https://user-images.githubusercontent.com/33585550/235757673-d6060056-fb9c-4525-a944-4c3c1a2bd21d.mov


philschatz (changes requested):
What do think of moving this code into the server? So far all filesystem access occurs in there (e.g. adding id attributes to all the elements).

TylerZeroMaster:
> What do think of moving this code into the server? So far all filesystem access occurs in there (e.g. adding id attributes to all the elements).

I was hoping to keep this self-contained since there is currently no overlap with functionality in the language server.

philschatz (approved):

PR #161, opened 2023-06-05
Add support for coderunner IFrame

philschatz:
This was the error in the VSCode terminal because the webview did not have permission to load local assets in the IFrame.

```
Refused to frame 'https://file+.vscode-resource.vscode-cdn.net/' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'frame-src' was not explicitly set, so 'default-src' is used as a fallback.
```

Refs https://github.com/openstax/ce/issues/2071

[Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src)


philschatz:
What seemed to be a one-line fix did not actually work.

PR #163, opened 2023-07-06
Full match license url

TylerZeroMaster:
Note to self: javascript's String.match is not a full match by default. Additionally, I should have had a test case for this. 🤦 

Additional known issue:
`this._licenseUrl.set(textWithRange(selectOne('/col:collection/col:metadata/md:license', doc), 'url'))`. When selectOne does not find anything, it throws an error, so the license url is never updated and retains its previous value. This means that the readme will be generated with the last valid values if the license element is removed altogether.

To me it makes more sense for the license url to be `null`  or `''` with a range pointing to nowhere if it is missing, but I think that change is outside of the scope of the readme-generator work.


server/src/model/cc-license.ts
```diff
@@ -14,7 +14,7 @@ const LICENSE_ATTRIBUTES: Record<string, string> = {
   sa: 'ShareAlike'
 }
 
-const LICENSE_PATTERN = /https?:\/\/creativecommons\.org\/licenses\/([a-z-]+)\/(\d\.\d)\/?(deed\..+)?/
+const LICENSE_PATTERN = /^https?:\/\/creativecommons\.org\/licenses\/([a-z-]+)\/(\d\.\d)\/?(deed\..+)?$/
```
sparksam, line 17: Can't get the regex to work with `https://creativecommons.org/licenses/by/4.0/aaa/` on [Regex101](https://regex101.com/)

sparksam: As long as it works in the tests. Then I guess it's fine by me.

TylerZeroMaster: The regex should not match `https://creativecommons.org/licenses/by/4.0/aaa/`. Prior to this change, that url would have matched because the regex did not include the beginning (`^`) and end (`$`) of the string: that was the mistake I made. Now the regex should not match unless the entire string is a valid license url.

therealmarv (approved):

omehes:
Some comments/test results:
Image 9-6-23 at 2 18 PM
Image 9-6-23 at 2 15 PM

in separate issue

PR #164, opened 2023-07-19
Submodule support

TylerZeroMaster:
These changes should prevent errors that could be caused by adding submodules to book repositories. These changes do not add support for committing and pushing changes to submodules.

- Stop assuming that the first repo is the book repo
- Get repo by pathHint (full path, partial path, or name)
- Set default git config on all repositories


ckline-tryptic (approved):

omehes:
testing: I opened otto-book repo, cloned tiny-book repo into it. Made changes in the tiny-book index file and pushed. Push failed - OK
push to open repo works

PR #169, opened 2023-09-20
2105 resource count quarx

TylerZeroMaster:
# Summary
Replaces excessive looping with a shared lists in the bundle for duplicate UUID and resource checks. These lists are automatically updated by quarx when the underlying models change.

Fixes openstax/ce#2105

# Duplicate Resources are Reported Correctly
## GIVEN
* [This](https://github.com/openstax/poet/suites/16545432679/artifacts/945128665) version of POET installed
* Any book repository open
## WHEN
* Two image names differ only by letter casing (i.e. `a.jpg` vs. `A.jpg`)
## THEN
* The images are highlighted red and a problem appears in the `Problems` pane

## GIVEN
* ... (same as above)
* Duplicate resource error reported by POET
## WHEN
* Two images that had names names differing only by letter casing are renamed to something different
## THEN
* The duplicate resource error goes away 

# Duplicate UUIDs are Reported Correctly
## GIVEN
* ...
## WHEN
* Two modules/pages have the same UUID
## THEN
* There is an error reported by POET

## GIVEN
* ...
* Duplicate UUID error reported by POET
## WHEN
* Two modules/pages that had the same UUID are updated to have different UUIDs
## THEN
* The duplicate UUID error goes away

### DIFF
--- a/server/src/model-manager.ts
+++ b/server/src/model-manager.ts
@@ -308,7 +308,7 @@ export class ModelManager {
       const stat = await fs.promises.stat(fsPath)
       if (stat.isFile()) { // Example: <image src=""/> resolves to 'modules/m123' which is a directory.
         return ['.jpg', '.png'].some((ext) => uri.endsWith(ext))
-          ? 'image'
+          ? '<fakeimagedata>'
           : await fs.promises.readFile(fsPath, 'utf-8')
       }
     }
--- a/server/src/model-manager.ts
+++ b/server/src/model-manager.ts
@@ -299,7 +299,9 @@ export class ModelManager {
     if (await checkFileExists(fsPath)) {
       const stat = await fs.promises.stat(fsPath)
       if (stat.isFile()) { // Example: <image src=""/> resolves to 'modules/m123' which is a directory.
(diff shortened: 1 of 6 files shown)


server/src/model-manager.ts
```diff
@@ -299,7 +299,9 @@ export class ModelManager {
     if (await checkFileExists(fsPath)) {
       const stat = await fs.promises.stat(fsPath)
       if (stat.isFile()) { // Example: <image src=""/> resolves to 'modules/m123' which is a directory.
-        return await fs.promises.readFile(fsPath, 'utf-8')
+        return ['.jpg', '.png'].some((ext) => uri.endsWith(ext))
+          ? 'image'
```
philschatz, line 303: Maybe something easier to grep like `<fakeimagedata>`

```suggestion
          ? '<fakeimagedata>'
```

philschatz (approved):

philschatz (approved):

omehes:
verified on ext dev-169 - discussed scenarios with Tyler
looks good

one bug found: https://github.com/openstax/ce/issues/2110 - also discussed

omehes:
verified on 6.5.2 - looks good

PR #171, opened 2023-09-26
adding slowing down tests and adding pages/home.py

omehes:


omehes:
Some locators/selectors are very long. They maybe could be made neater but I couldn't find a way (so far)...

TylerZeroMaster (approved):
Wow, those are some lengthy selectors; but if they work then I do not see a problem, personally.

omehes:
@TylerZeroMaster Yeah, that is the thing, that if I tried to shorten some of the selectors, locators wouldn't work. But from Mike's times, I don't like long and complicated selectors... Anyway, I will try to amend them, If I can...

PR #174, opened 2023-10-04
adding qa test to verify problems tab in poet

omehes:
Test checks that each book repo Problems tab open in gitpod has 0 probs
Image 10-4-23 at 8 15 AM

### DIFF
--- a/tests/ui/test_problems_tab_all_repos.py
+++ b/tests/ui/test_problems_tab_all_repos.py
@@ -1,0 +1,103 @@
+import pytest
+
+from tests.ui.pages.home import HomePoet
+
+from bs4 import BeautifulSoup
+
+
+@pytest.mark.nondestructive
+def test_problems_tab_all_repos(
+    chrome_page, github_user, github_password, git_content_repos, headers_data
+):
+    # GIVEN: Playwright, chromium and gitpod
+    # repo url as: https://gitpod.io/#https://github.com/openstax/osbooks-otto-book
+    # run the test: pytest -k test_problems_tab.py tests/ui --github_user xxx --github_password yyy --github_token zzz
+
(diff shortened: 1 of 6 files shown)


omehes:
I am still investigating random ```playwright._impl._api_types.Error: Element is not attached to the DOM``` error but I would appreciate any feedback to the test/code so far...

tests/ui/test_problems_tab_all_repos.py
```diff
@@ -0,0 +1,103 @@
...
+
+                ccont = chrome_page.content()
+
+                sopa = BeautifulSoup(ccont, "html.parser")
+                problems = sopa.select(aria_label_0)
+
+                repo_name = gitpod_repo_url.split("openstax/")
+
+                try:
+                    assert problems
```
TylerZeroMaster, line 93: In my opinion, it would make more sense if this was `assert no_problems` or `assert zero_problems` since you are checking if the `Problems (⇧⌘M) - Total 0 Problems` element exists.

tests/ui/test_problems_tab_all_repos.py
```diff
@@ -0,0 +1,103 @@
...
```
TylerZeroMaster, line 50: See my other comment about this.

TylerZeroMaster (changes requested):

TylerZeroMaster (approved):
This will still report problems that are info and/or warning level. You could try filtering those out and checking the filtered count instead if you want to avoid this. For example, 56 of these problems are errors:
Screenshot 2023-10-05 at 12 24 49 PM

Granted, this works and I am not sure if you want to filter out the info/warnings, so I feel like it should be approved 👍

PR #177, opened 2023-10-11
page object selectors cleanup/simplification

omehes:


TylerZeroMaster (approved):
`div[class="<class-name>"]` should be function the same as `div.<class-name>` as long as there is only one class. If there is more than one class, `div[class="<class-name> <class-2>"]` only matches elements that have the classes specified **in that exact order** whereas `div.<class-name>.<class-2>` is the same as `div.<class-2>.<class-name>`.

PR #178, opened 2023-10-19
removed check for workspace limit

omehes:


ckline-tryptic (approved):
The autoformatting is blinding :wink: but I don't see any problems with the code.

PR #181, opened 2023-10-30
adding new qa test to check new poet buttons (readme, validate)

omehes:


tests/ui/test_readme_validate_other_features.py
```diff
@@ -0,0 +1,59 @@
...
+
+    home.click_generate_readme_button()
+    assert home.generate_readme_dialog_box_is_visible
+    assert "Generate README: Done!" in home.generate_readme_dialog_box_text.inner_text()
+
+    home.click_validate_content_button()
+    assert home.validate_content_popup_dialog_is_visible
+
+    home.click_validate_content_cancel_button()
+    assert not home.validate_content_popup_dialog_is_visible
```
TylerZeroMaster, line 56: You might need to wait for the popup to no longer be visible. When you click cancel, it can take a few seconds for the operation to stop and for the popup to disappear. Although, it might be okay since you have a global slowdown set.

TylerZeroMaster (approved):

PR #185, opened 2023-11-14
:bug: fix dangling comment node

philschatz:
This was causing the preview to fail when a comment occurred after the root `<document>` element.

refs https://github.com/openstax/ce/issues/2129


TylerZeroMaster (approved):
Looks good. If I understand correctly, comments were being parsed as document fragments when they were at the root level. Since comments are not handled, they became null and created the issue.

therealmarv (approved):

PR #186, opened 2023-11-30
amending validation test as per gitpod changes

omehes:
and adding a quick code to get indexes of all content repos in case I want to devide the test into smaller chunks


TylerZeroMaster (approved):
Thank you for removing the old version. This way it is also easier to compare changes since GitHub shows individual changes instead of entirely new file. 

Looks like it the login process is a bit more tricky to automate now. This looks good, but I wonder if it is work trying to use some of the more general playwright locators, like [get_by_text](https://playwright.dev/python/docs/api/class-page#page-get-by-text), in the future. I think something like `self.page.locator("button").get_by_text("Continue with GitHub")` could potentially work without relying on very specific CSS selectors. Granted, that would break if the text of the button changed. It is hard to say which way is better.

PR #188, opened 2023-12-05
:art: update GitHub actions and node version

philschatz:
It's been a while since we updated these and GitHub [started warning about them](https://github.com/openstax/poet/actions/runs/7065738946):

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-node@v2, actions/upload-artifact@v2, codecov/codecov-action@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


TylerZeroMaster (changes requested):
Looks good except that I think code coverage from cypress tests is broken. When I run `test:cypress`, it does not generate any coverage in `.nyc_output` or `coverage` directories.

According to [this documentation](https://docs.cypress.io/guides/tooling/code-coverage#Install-the-plugin), the way we install the coverage plugin may be outdated. I tried removing `cypress/plugins/index.ts` and replacing `cypress.config.js` with this:

```typescript
import { defineConfig } from 'cypress'

export default defineConfig({
  e2e: {
    baseUrl: null,
    setupNodeEvents(on, config) {
      require('@cypress/code-coverage/task')(on, config)
      return config
    },
  }
})
```

That seemed to fix the issue.

philschatz:
Whoops, I was just looking at the 100% & missed that `clinet/src/webview-js/**` was dropped :blush: . Fixed!

PR #189, opened 2024-01-26
new test-tool to verify .github/settings.yml of each book repo

omehes:


tests/ui/content/test_settings_yml_all_book_repos.py
```diff
@@ -0,0 +1,65 @@
...
+                        f"Error code {settings_yml_resp.status_code}: Incorrect/missing settings.yml file"
+                    )
+                    continue
+
+                else:
+                    settings_yml_content = yaml.safe_load(settings_yml_resp.text)
+                    settings_yml_list = list(settings_yml_content.values())[0]
+
+                    try:
+                        settings_yml_list["private"]
```
TylerZeroMaster, line 49: This seems a little strange to me. The structure of the settings.yml should be consistent, but the order of properties is not. I think it would make more sense, possibly, to do this:
```python
try:
    is_private = settings_yml_content["repository"]["private"]
    if is_private != False:
        print(f"repository:private entry in .github/settings.yml is set to {is_private} of type {type(is_private)}")
        continue
except KeyError:
    print("!!! issue with private key in .github/settings.yml")
```
This way the structure of the yaml is tested and the type of the value as well (hence `is_private != False` instead of `not is_private`).

Also keep in mind that your current implementation assumes `private` will always be the first entry of the first value of the yaml. While that might be true now, it might not always be the case since `private` could be anywhere in the list.

TylerZeroMaster (changes requested):

tests/ui/content/test_settings_yml_all_book_repos.py
omehes: I see... didn't think about that...

omehes:
@TylerZeroMaster changes done. thanks

tests/ui/content/test_settings_yml_all_book_repos.py
```diff
@@ -0,0 +1,64 @@
...
+                    )
+                    continue
+
+                else:
+                    settings_yml_content = yaml.safe_load(settings_yml_resp.text)
+
+                    try:
+                        settings_yml_list = list(settings_yml_content.values())[0]
+
+                    except AttributeError:
```
TylerZeroMaster, line 50: I think this should be an `IndexError` if you are converting `settings_yml_content` to list and then indexing that list to determine if it has any values.

tests/ui/content/test_settings_yml_all_book_repos.py
```diff
@@ -0,0 +1,64 @@
...
+                        try:
+                            is_private = settings_yml_content["repository"]["private"]
+
+                            if is_private:
+                                print(
+                                    f"<<< repository:private entry in .github/settings.yml is set to {is_private}"
+                                )
+                                continue
+
+                        except (KeyError, TypeError):
```
TylerZeroMaster, line 63: This `TypeError` will never happen.

TylerZeroMaster (changes requested):

tests/ui/content/test_settings_yml_all_book_repos.py
omehes: It happend, that is why I eventually added it (I think I had only key but no value to it)...

tests/ui/content/test_settings_yml_all_book_repos.py
omehes: If for example settings.yml is empty, I get AttributeError here, so I have to leave it

omehes:
@TylerZeroMaster One change ready

TylerZeroMaster (approved):

PR #191, opened 2024-02-14
test to verify .gitpod.yml of each book repo

omehes:


omehes:
@TylerZeroMaster Thanks. I like your solution. Looks neat.

PR #192, opened 2024-02-19
2165 push to private submodule

TylerZeroMaster:
fixes openstax/ce#2165

# Changes are backward compatible 
## GIVEN
* A book repository that does not have a private submodule
* [This](https://github.com/openstax/poet/actions/runs/8024151015/artifacts/1270765692) version of POET
## WHEN
* POET is activated (by opening its interface or opening a cnxml document)
## THEN
* It works as usual (private submodule is not required)
* Can push changes
* No errors on startup

# Push to private submodule and book repo
**NOTE**: It may be best to use the [private-submodule](https://github.com/openstax/osbooks-playground/tree/private-submodule) branch of osbooks-playground which has a private submodule preconfigured and ready to go.
## GIVEN
* A book repository contains a properly configured private submodule is opened in gitpod
* [This](https://github.com/openstax/poet/actions/runs/8024151015/artifacts/1270765692) version of POET
## WHEN (order is important)
1. You have read and write access to the private submodule
2. POET is activated (by opening its interface or opening a cnxml document)
3. You make a change to a file in the private submodule
## THEN
* The private submodule is checked out to the correct branch (defined in the .gitmodules file)

    * You can check this by running `git -C private branch --show-current` in the terminal

Screenshot 2024-02-27 at 11 30 48 AM

* You can push changes to public and private submodule

# Making private submodule changes before POET is activted
**NOTE**: It may be best to use the [private-submodule](https://github.com/openstax/osbooks-playground/tree/private-submodule) branch of osbooks-playground which has a private submodule preconfigured and ready to go.
## GIVEN
* A book repository contains a properly configured private submodule is opened in gitpod
* [This](https://github.com/openstax/poet/actions/runs/8024151015/artifacts/1270765692) version of POET
## WHEN (order is important)
1. You have read and write access to the private submodule
2. You make a change to a file in the private submodule
3. POET is activated (by opening its interface or opening a cnxml document)
## THEN
* The private submodule is checked out to the correct branch (defined in the .gitmodules file)
    * You can check this by running `git -C private branch --show-current` in the terminal or looking in the git SCM panel

Screenshot 2024-02-27 at 11 30 48 AM

* You can push changes to public and private submodule

# Error message when submodule is not initialized
**BACKGROUND**: These steps are supposed to create an uninitialized submodule in gitpod. Normally when you open gitpod, it does something like `git clone --recursive` which clones the repository and all submodules. When you start on a branch that does not contain submodules and then switch to one that does, however, the submodules are not cloned/initialized.
## GIVEN
* [no-private-submodule](https://github.com/openstax/osbooks-playground/tree/no-private-submodule) branch of osbooks-playground opened in gitpod
* [This](https://github.com/openstax/poet/actions/runs/8024151015/artifacts/1270765692) version of POET
## WHEN (order is important)
1. You switch to the private-submodule branch of osbooks-playground
2. POET is activated (by opening its interface or opening a cnxml document)
## THEN
* An error message appears warning you that your changes will not be pushed because the private submodule is not initialized
* Trying to push causes the error message to appear again
* If there are changes in the book repository, these are still pushed

---

# Summary of Changes

Unfortunately, the git extension [does not surface](https://github.com/microsoft/vscode/blob/7215958b3c57945b49d3b70afdba7fb47319ca85/extensions/git/src/api/git.d.ts#L65) the `branch` property of `.gitmodules`. It [parses the file](https://github.com/microsoft/vscode/blob/7215958b3c57945b49d3b70afdba7fb47319ca85/extensions/git/src/git.ts#L739) in nearly the exact same way, but [branch is omitted](https://github.com/microsoft/vscode/blob/7215958b3c57945b49d3b70afdba7fb47319ca85/extensions/git/src/git.ts#L854) for some reason.

The only alternative was to try to guess the edition of the book and construct the branch name using the guessed edition and workspace root directory name. Parsing the configuration seemed like the better of the two options since it makes fewer assumptions and is less reliant on repository file structure.

Making a PR to vscode to include the branch might be a good thing to do in the long-term.

### DIFF
--- a/client/src/push-content.ts
+++ b/client/src/push-content.ts
@@ -272,7 +272,9 @@ export const pushContent = (hostContext: ExtensionHostContext) => async () => {
       if (privateSubmodule !== undefined) {
         try {
           await preparePrivateSubmodule(hostContext.client, privateSubmodule)
-          pushTargets.push(privateSubmodule)
+          // Private submodule should go first because that means the latest
+          // submodule version will be committed
+          pushTargets.unshift(privateSubmodule)
         } catch (e) {
           const err = e as Error
           await vscode.window.showErrorMessage(err.message)
--- a/client/src/push-content.ts
+++ b/client/src/push-content.ts
@@ -210,16 +210,33 @@ export const getMessage = async (): Promise<string | undefined> => {
   return message
 }
 
+export const _hasSubmodule = (repo: Repository, name: string) => {
+  return repo.state.submodules.some((mod) => mod.name === name)
+}
+
 // NOTE: When the private submodule is not initialized, this will return undefined
-export const _getPrivateSubmodule = () => getRepo(PRIVATE_SUBMODULE_NAME)
+export const _getPrivateSubmodule = () => {
+  const privateSubmodule = getRepo(PRIVATE_SUBMODULE_NAME)
+  if (privateSubmodule === undefined) {
+    /* istanbul ignore else (should only print a warning) */
+    if (_hasSubmodule(getBookRepo(), PRIVATE_SUBMODULE_NAME)) {
+      // TODO: Wish we could initialize submodules when this happens
+      throw new Error(
+        'ERROR: Uninitialized private submodule detected! ' +
+        'Your changes will not be committed/pushed!'
+      )
+    } else {
+      console.warn(`Private submodule not found: ${PRIVATE_SUBMODULE_NAME}`)
+    }
+  }
+  return privateSubmodule
+}
 
 /* istanbul ignore next (already tested in pushContent) */
 export const initPrivateSubmodule = async (hostContext: ExtensionHostContext) => {
   const privateSubmodule = _getPrivateSubmodule()
   if (privateSubmodule !== undefined) {
     await preparePrivateSubmodule(hostContext.client, privateSubmodule)
-  } else {
-    console.warn(`Private submodule not found: ${PRIVATE_SUBMODULE_NAME}`)
   }
 }
 
@@ -227,7 +244,6 @@ export const preparePrivateSubmodule = async (
   client: LanguageClient,
   privateSubmodule: Repository
 ): Promise<void> => {
-  // TODO: What if submodule needs to be initialized? (not a problem in gitpod)
   const uri = expect(getRootPathUri(), 'Could not get root path')
   const maybeGitModules = await requestGetSubmoduleConfig(
     client,
@@ -280,8 +296,6 @@ export const pushContent = (hostContext: ExtensionHostContext) => async () => {
           await vscode.window.showErrorMessage(err.message)
           return
         }
-      } else {
-        console.warn('Could not find private submodule')
       }
       // push content
       progress.report({ message: 'Pushing...' })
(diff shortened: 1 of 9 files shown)


client/src/push-content.ts
```diff
@@ -207,6 +210,44 @@ export const getMessage = async (): Promise<string | undefined> => {
...
+  } else {
+    console.warn(`Private submodule not found: ${PRIVATE_SUBMODULE_NAME}`)
+  }
+}
+
+export const preparePrivateSubmodule = async (
+  client: LanguageClient,
+  privateSubmodule: Repository
+): Promise<void> => {
+  // TODO: What if submodule needs to be initialized? (not a problem in gitpod)
```
philschatz, line 230: I'm probably just confused but doesn't this code handle this `TODO`?

philschatz (approved):
Nice!

server/src/server-handler.ts
```diff
@@ -35,6 +36,15 @@ export function bundleGenerateReadme(): (request: BundleGenerateReadme) => Promi
   }
 }
 
+export function bundleGetSubmoduleConfig(): (request: BundleGetSubmoduleConfigParams) => Promise<Record<string, string> | null> {
```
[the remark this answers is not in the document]
TylerZeroMaster: Yes, 💯

client/src/push-content.ts
TylerZeroMaster: No, initially when clone a repository that contains submodules, the submodules are not initialized. They do not contain a `.git` directory until you run something like `git submodule update --init`. Since they do not contain a `.git` directory, they are not considered git repositories and git commands like `fetch` fail.

Gitpod automatically initializes submodules when you open a workspace. This TODO is specifically for making it work locally. I was exploring ways to make it work. One possible way to get it to work right now would be to use the `git.clone` command. Since it works in Gitpod and you can workaround the issue locally with by initializing submodules manually, it seemed like something that could wait until later.

server/src/git-config-parser.ts
```diff
@@ -0,0 +1,28 @@
+// RegExp from: vscode/extensions/git/src/git.ts
+const lineSep = /\r?\n/
+const propertyPattern = /^\s*(\w+)\s*=\s*"?([^"]+)"?$/
+const sectionPattern = /^\s*\[\s*([^\]]+?)\s*("[^"]+")*\]\s*$/
+
+export function parseGitConfig(config: string): Record<string, string> {
```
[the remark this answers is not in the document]
TylerZeroMaster: @philschatz I share your frustration. The only other option I could surmise was to try to reconstruct the branch name based on the repository directory name and the currently checked out branch. That approach seemed more likely to cause unexpected issues later on. Moreover, I thought it was be better to use an existing git standard than to try to develop a proprietary one.

client/src/push-content.ts
TylerZeroMaster: I tried making this more clear by adding an error that is thrown if the private submodule is unexpectedly missing. This would happen when the submodule is defined in the `.gitmodules` and:
1. The directory is missing
2. The directory is empty
3. Any other situation where the git extension does not think it is a valid git repository

omehes:
verified in poet ext 7.0.0

PR #193, opened 2024-03-08
Use modal error for submodule errors during push

TylerZeroMaster:
Some changes suggested by @omehes to improve user experience:

When changes cannot be pushed to the private submodule:
  - A modal error dialog displays the error
  - There is an option to continue pushing changes to the book repository

Prior to this change, errors in preparing the private submodule would cause the entire push process to be aborted. Now there is an option to continue or bail out

Screenshot 2024-03-08 at 9 22 50 AM

Screenshot 2024-03-08 at 10 07 30 AM

### DIFF
--- a/client/src/push-content.ts
+++ b/client/src/push-content.ts
@@ -221,10 +221,7 @@ export const _getPrivateSubmodule = () => {
     /* istanbul ignore else (should only print a warning) */
     if (_hasSubmodule(getBookRepo(), PRIVATE_SUBMODULE_NAME)) {
       // TODO: Wish we could initialize submodules when this happens
-      throw new Error(
-        'ERROR: Uninitialized private submodule detected! ' +
-        'Your changes will not be committed/pushed!'
-      )
+      throw new Error('Uninitialized private submodule detected!')
     } else {
       console.warn(`Private submodule not found: ${PRIVATE_SUBMODULE_NAME}`)
     }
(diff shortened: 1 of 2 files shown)


client/src/push-content.ts
```diff
@@ -284,16 +281,24 @@ export const pushContent = (hostContext: ExtensionHostContext) => async () => {
...
-          // submodule version will be committed
+          // submodule version will be committed to the book repository
           pushTargets.unshift(privateSubmodule)
-        } catch (e) {
-          const err = e as Error
-          await vscode.window.showErrorMessage(err.message)
+        }
+      } catch (e) {
+        const err = e as Error
+        // Add '.' if the message does not end in '.', '?', or '!'.
```
philschatz, line 294: :smile:

client/src/push-content.ts
```diff
@@ -284,16 +281,24 @@ export const pushContent = (hostContext: ExtensionHostContext) => async () => {
...
+        const errMsg = err.message.trimEnd().replace(/[^.?!]$/, (s) => `${s}.`)
+        const response = await vscode.window.showErrorMessage(
+          `${errMsg} Your changes will not be committed/pushed to the private submodule!`,
```
philschatz, line 297: Could append it to 'Your changes will not..." rather than prepending to that string? Maybe something like:

```suggestion
          `Your changes will not be committed/pushed to the private submodule!\nUnderlying reason:\n${errMsg}`,
```

philschatz (approved):
Just a small idea to avoid parsing english sentences in error messages. Also helps clarify that the error message isn't coming from POET directly (so searching for it in the repo won't help)

client/src/push-content.ts
TylerZeroMaster: Hmm, putting the conclusion at the beginning seems strange to me, but I can see your point about how it could cause problems with localization.

PR #194, opened 2024-05-01
2209 h5p exercise url

TylerZeroMaster:
fixes openstax/ce#2209

# Summary

- Add H5P exercise type (not part of resources because of the number of assumptions made about resources).
- Add note about paths being hardcoded into file system watchers and likelihood of watcher system changing.
- Check H5P exercise exists as part of page validation.
- Add H5P exercise url autocompletion.

# Known issues

While making these changes, I found a few problems and I am working on making new issues for each since they are outside of the scope of 2209.

- Directory creation events are not handled at all. Moving a directory containing content that POET understands into an area POET is watching does not result in that content being added to the bundle.
- Page validation errors related to content not existing, i.e. resource or h5p, do not update until the page updates.
- Content paths are hardcoded into client where file system watchers are created.

### DIFF
--- a/client/src/utils.ts
+++ b/client/src/utils.ts
@@ -165,10 +165,20 @@ export function launchLanguageServer(context: vscode.ExtensionContext): Language
     documentSelector: [{ scheme: 'file', language: 'xml' }],
     synchronize: {
       fileEvents: [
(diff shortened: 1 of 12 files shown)


philschatz (approved):
:+1:

PR #195, opened 2024-05-10
Mark nodes removed instead of deleting them

TylerZeroMaster:
fixes openstax/ce#2217

# Problem:

- When POET loads a page, references to resources on the page are stored in the  page node (see _resourceLinks)

- When a resource was removed from the bundle, perhaps via a filesystem event, the references in the page remained, even though the resources no longer existed in the bundle.

- When a resource with the same absPath was created, it was added to the bundle again. This new resource instance, however, was distinct from the one that existed within the page, even though they pointed to the same content on the filesystem.

- Until the page was updated, which causes the links to be reconstructed, resource links stored in the page did not correlate with the ones stored in the bundle. Consequently, the `exists` property of the resource node was not being updated

# Solution:

- Keep all nodes in the bundle

- When a file referenced by a node is deleted from the filesystem, update the node's contents to `undefined` so that the node's `exists` property will be set to false

# Possible problems & justifications:

- This technically introduces a memory leak where nodes continue to accumulate in the bundle forever; however, there are many ways this memory leak could be addressed if it becomes a real problem.

- Out of all the approaches I explored, this one made the best use of Quarx, required the least number of changes, and exhibited the best performance.

### DIFF
--- a/server/src/model-manager.spec.ts
+++ b/server/src/model-manager.spec.ts
@@ -317,6 +317,8 @@ describe('processFilesystemChange()', () => {
     // Load the Bundle, Book, and Page
     loadSuccess(first(loadSuccess(first(loadSuccess(manager.bundle).books)).pages))
 
+    const sizeBefore = manager.bundle.allNodes.size
+
     // Delete non-existent file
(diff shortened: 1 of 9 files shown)


philschatz (approved):
:+1:

PR #196, opened 2024-05-15
2216 handle directory creation

TylerZeroMaster:
part of openstax/ce#2216

Directory events are separate from file events. When a directory is created, that results in a single event for the directory itself. For poet, this can create issues because copying a module directory, for example, will not result in a new page node being created.


philschatz (approved):
:+1:

therealmarv (approved):

PR #203, opened 2024-06-24
More validation for assumptions made in Enki

TylerZeroMaster:
Book slugs are defined in three places--books.xml, collection file name, collection file metadata--and must match in all three places. This is a common source of build errors in newly created books. Most processes in Enki use the slug from the books.xml file, however, there are a few exceptions. 

Codify the assumptions made about book slugs.

- Validate that the slug in the books.xml matches the one in the associated collection metadata (bakery-scripts)
- Validate that the collection name matches what epub builds would expect (bakery-js and CORGI)


philschatz (approved):
Nice! :+1:

PR #204, opened 2024-06-25
Add Ancillary feature

sparksam:
Ref: https://github.com/openstax/ce/issues/2223 
Ref: https://github.com/openstax/ce/issues/1910
Ref: https://github.com/openstax/ce/issues/2230

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -16,7 +16,7 @@
         "@types/node": "^17.0.16",
         "@types/react-sortable-tree": "^0.3.15",
         "@types/uuid": "^8.3.4",
-        "@types/vscode": "^1.64.0",
+        "@types/vscode": "^1.89.0",
         "@types/xmldom": "^0.1.31",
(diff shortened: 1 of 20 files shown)


TylerZeroMaster (changes requested):
Looks good, but there are some things not covered by tests. If I look at the html report (in `.nyc_output/lcov-report` after running `npm run test:unit`) it looks like most of the uncovered parts are where the functions return early if no title is entered. 

In my opinion, it's fine to either nest the logic that expects a title inside the if statement instead of returning early, or istanbul ignore the if statements that check the title (since they only return, I do not see it being a big deal to ignore them).

Also, the validation function for `askTitle` is not covered by tests. I would maybe pull this function out of the `askTitle` function and make it a new function that you can test separately. Alternatively, you could use a mock implementation of `vscode.window.showInputBox` that calls the validation function.

TylerZeroMaster (approved):

PR #207, opened 2024-07-01
Upgrade to mathml3

TylerZeroMaster:
see openstax/ce#1584 if you would like additional background information
related to openstax/ce#1909
related to openstax/cnxml#65

Unfortunately a patch is required to add the correct math elements to two definitions of `semantics`.


sparksam (approved):
Looks good to me.

PR #208, opened 2024-07-17
Slug uri compare fix

TylerZeroMaster:
Use `basename` to verify collection is named as expected 

The problem was that some characters were being uri encoded on one side of the comparison but not the other

Using basename on both side of the comparison should cause both strings to undergo the same transformation

As long as both results are the same, then the validation should pass


philschatz (approved):
lgtm!

PR #215, opened 2025-01-02
Core 617 orphan collection

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-617

Screenshot 2025-01-02 at 2 09 30 PM

I also fixed a problem where ancillary tree items were being rendered with the same icon as a subbook despite being closer to pages.

### DIFF
--- a/client/src/tocs-event-handler.ts
+++ b/client/src/tocs-event-handler.ts
@@ -1,5 +1,5 @@
 import vscode from 'vscode'
-import { type TocsTreeProvider, type BookOrTocNode } from './book-tocs'
+import { type TocsTreeProvider, type BookOrTocNode, ClientOnlyTocKinds } from './book-tocs'
 import { type TocModification, TocModificationKind, type TocModificationParams, TocNodeKind, BookRootNode, type CreatePageEvent, type CreateSubbookEvent, type CreateAncillaryEvent, isClientTocNode } from '../../common/src/toc'
 import { ExtensionServerRequest } from '../../common/src/requests'
 import { expect, getRootPathUri } from './utils'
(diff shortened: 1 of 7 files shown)


client/src/tocs-event-handler.ts
```diff
@@ -204,10 +208,18 @@ export class TocsEventHandler implements vscode.TreeDragAndDropController<BookOr
...
-      await this.moveNode(dragging, target)
+    if (dragging?.type === undefined) { throw new Error('BUG: Bad drag target') }
+    if (target === undefined) { throw new Error('BUG: Bad drop target') }
+    if (
+      target !== dragging &&
+      dragging.type !== ClientOnlyTocKinds.OrphanCollection &&
+      dragging.type !== BookRootNode.Singleton
+    ) {
+      if (target.type === ClientOnlyTocKinds.OrphanCollection) {
+        await this.removeNode(dragging)
```
Dantemss, line 219: So they never get put in the OrphanCollection, they just go under it somehow? How does the OrphanCollection actually gets elements, like in your screenshot?

Dantemss (commented):
My first time looking at this codebase, but I'll give it a shot

client/src/tocs-event-handler.ts
TylerZeroMaster: - Orphans are sent by the language server [here](https://github.com/openstax/poet/blob/30dcb65f7a3449418ef89e3aeb3e7141065c2d56/server/src/model-manager.ts#L173) after the remove event is handled.
- [This](https://github.com/openstax/poet/blob/de8a57e6e4827904686999c21a9767cf945d8db2/client/src/extension.ts#L102) updates the client's toc tree with books and orphans.
- Orphans are added to the OrphanCollection [here](https://github.com/openstax/poet/blob/4bdf7c59fef017cf3bfa929778b2b5d32f1537a1/client/src/book-tocs.ts#L47) on update.

client/src/tocs-event-handler.ts
```diff
@@ -126,41 +128,43 @@ export class TocsEventHandler implements vscode.TreeDragAndDropController<BookOr
...
   }
 
   async renameNode(node: BookOrTocNode) {
     // TODO Implement the rename functionality using inline editing (wait for the API to be available)
     // https://github.com/microsoft/vscode/issues/97190
     // https://stackoverflow.com/questions/70594061/change-an-existing-label-name-in-tree-view-vscode-extension
     let oldTitle: string | undefined = ''
-    if (node.type !== BookRootNode.Singleton && 'title' in node.value) {
-      /* istanbul ignore next */
+    if (isClientTocNode(node)) {
```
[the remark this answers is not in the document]
TylerZeroMaster: I agree that this was a little confusing. My goal was to illustrate that the orphan collection is something that is specific to the tree view. I did not want to entangle it with the `ClientTocNode`. I replaced the `ClientOnlyTocKinds` enum with an `OrphanCollectionKind` constant in hopes of making the difference more clear.

client/src/tocs-event-handler.ts
TylerZeroMaster: Also, here's a video instead of a screenshot. 

https://github.com/user-attachments/assets/17d6f173-3b70-4b39-8a85-a80353e2ddb7

Dantemss (approved):
Thanks for the clarifications, and the change looks good to me.

